PR #67, opened 2015-10-20
Sa/update docstrings

genestack-solomatin:
- fix links to python docs used in API reference
- break long lines
- small cleanups

### DIFF
--- a/docs/source/create_cla_docs.py
+++ b/docs/source/create_cla_docs.py
@@ -19,7 +19,7 @@ print os.getcwd()
 
 
 def main():
-    with open('cla_reference._rst', 'w') as f:
+    with open('_cla_reference.rst', 'w') as f:
         for name, attr in getmembers(cla):
             if isclass(attr) and name != 'CLApplication' and issubclass(attr, cla.CLApplication):
                 f.write(template.format(name=name, name_underline='-' * len(name)))
--- a/docs/source/create_cla_docs.py
+++ b/docs/source/create_cla_docs.py
@@ -1,0 +1,25 @@
+from inspect import getmembers, isclass
+from genestack_client import cla
+
+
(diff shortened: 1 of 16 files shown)


docs/source/create_cla_docs.py
```diff
@@ -0,0 +1,25 @@
...
+        .. autoattribute:: genestack_client.{name}.APPLICATION_ID
+
+'''
+
+import os
+print os.getcwd()
+
+
+def main():
+    with open('cla_reference._rst', 'w') as f:
```
daniel-ohayon: why the ``._rst`` extension? 
If this was intentional, add ``cla_reference._rst`` to the ``.gitignore``

genestack-solomatin: It was done to remove `WARNING: document isn't included in any toctree` but I found way to do it via `excludes` in `conf.py`

daniel-ohayon:
the sphinx syntax for python classes is not used everywhere consistently in the docs.
For example, sometimes `argparse.ArgumentParser` is written as is, and sometimes it is written as `:py:class:`~argparse.ArgumentParser``

daniel-ohayon:
Another docs-related issue is that the return type of `find_files_by_name` is documented as `IFile`, which is false.

genestack-solomatin:
``:py:class:`~argparse.ArgumentParser` `` should be used inside text entries and  ``argparse.ArgumentParser`` for `:rtype:`

genestack-solomatin:
TODO for next prs:
- merge https://github.com/genestack/python-client/tree/vb/update_documentation (after merge of this branch)
- add utils to api reference 
- rearrange scripts and docs in source folder and .gitignore files (make one special for docs folder)

PR #69, opened 2015-10-21
move rename_file to files util from CLA

Juhlchen:


genestack_client/files_util.py
```diff
@@ -468,3 +468,18 @@ def get_infos(self, accession_list):
...
+    def rename_file(self, accession, name):
+        """
+        Rename a file and returns its new name.
+
+        :param accession: file accession
+        :type accession: str
+        :param name: name
+        :type name: str
+        :return: new name
+        :rtype: str
```
genestack-solomatin: `:rtype:` should match with return type

genestack_client/files_util.py
```diff
@@ -468,3 +468,18 @@ def get_infos(self, accession_list):
...
+        """
+        # TODO java returns file info because javascript require it for callback, should we support same interface in python?
```
genestack-solomatin: I think we should remove `return`, it does not give us any new information

genestack_client/cla.py
```diff
@@ -129,21 +129,6 @@ def start(self, accession):
         """
         self.invoke('start', accession)
 
-    # TODO move to filesUtils, why we return name?
-    def rename_file(self, accession, name):
```
genestack-solomatin: We need to return this method.  User need to have some time to fix theirs scripts. 

Implementation should be:
```python
# depricated at 25 remove at 26
sys.stderr.write('Depicated ...  use ... instead\n')
FilesUtil(self.connection).rename_file(app_file, name)
```
Current dotorg release is 24, this method will be present while release 25 is on dotorg and will not work after release 26

genestack-solomatin:
I talked about our workflow, now master should be compatible only with development server.

PR #70, opened 2015-10-26
Add log of error in jar file

genestack-solomatin:
Log file name in case of any other errors. For example `BadZipfile`.

### DIFF
--- a/genestack-application-manager
+++ b/genestack-application-manager
@@ -394,28 +394,41 @@ AppInfo = namedtuple('AppInfo', [
 ])
 
 
-def read_jar_file(file_path):
-    try:
-        with zipfile.ZipFile(file_path) as zip_file:
-            try:
-                info = zip_file.getinfo('applications.xml')
-                with zip_file.open(info) as manifest:
-                    doc = minidom.parse(manifest)
-                    namespace = doc.documentElement.namespaceURI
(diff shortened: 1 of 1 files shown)


genestack-application-manager
```diff
@@ -395,22 +395,27 @@ AppInfo = namedtuple('AppInfo', [
...
-                vendor = doc.getElementsByTagNameNS(namespace, 'vendor')[0].firstChild.nodeValue
-                identifiers = [
-                    vendor + '/' + a.getElementsByTagNameNS(namespace, 'id')[0].firstChild.nodeValue
-                    for a in applications
-                ]
-        except KeyError as e:
-            raise GenestackException('Unable to read applications.xml manifest from %s: %s' % (
-                os.path.abspath(file_path), e))
-        return AppInfo(vendor, identifiers)
+    try:
```
klas-genestack: Wrapping the whole block into `try: .. except` is bad, I'd use [testzip()](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.testzip) for check file before doing anything with it.

genestack-solomatin:
If file is not present or not `zip` archive exception will be throw in constructor, testzip() will not help. 
I want to print file path in case of any error.

klas-genestack:
Still *five* levels of indentation is a problem.

PR #75, opened 2015-11-09
Added prompting for actions which will remove system stable version o…

ekozobrodov:
…f application;

Added flag 'force' for 'install' and 'remove' command.

### DIFF
--- a/genestack-application-manager
+++ b/genestack-application-manager
@@ -18,6 +18,7 @@ import json
 from collections import namedtuple
 from genestack_client import GenestackException
 from genestack_client.genestack_shell import GenestackShell, Command
+from genestack_client.utils import isatty
 
 
 def validate_application_id(app_id):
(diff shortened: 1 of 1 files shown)


genestack-application-manager
```diff
@@ -438,6 +461,25 @@ def show_info(files, vendor_only, with_filename, no_filename):
...
 
 
+def prompt_removing_stable_version(application, apps_ids, version):
+    apps = []
+    for app_id in apps_ids:
+            stable_versions = application.invoke('getStableVersions', app_id)
+            if stable_versions.get('SYSTEM') == version:
+                apps.append(app_id)
+
+    if len(apps) == 0:
```
genestack-solomatin: `if not apps:`

genestack-application-manager
```diff
@@ -337,18 +354,26 @@ def reload_applications(application, version, app_id_list):
...
 
 
 def upload_single_file(application, file_path, version, override,
-                       stable, scope):
+                       stable, scope, force=False):
+    app_info = read_jar_file(file_path)
+    if not force and override and not (stable and SCOPE_DICT[scope] == 'SYSTEM'):
+        if not prompt_removing_stable_version(application, app_info.identifiers, version):
+            sys.stdout.flush()
+            sys.stderr.write('Installation was aborted by user\n')
```
genestack-solomatin: It is not true, files were installed.  Only set stable with scope USER is aborted

genestack-application-manager
```diff
@@ -438,6 +461,25 @@ def show_info(files, vendor_only, with_filename, no_filename):
...
+    for app_id in apps_ids:
+            stable_versions = application.invoke('getStableVersions', app_id)
+            if stable_versions.get('SYSTEM') == version:
+                apps.append(app_id)
+
+    if len(apps) == 0:
+        return True
+
+    message = ('You are going to make unstable following system stable applications with version "%s": \n' % version) +\
+        '\n'.join(apps) + "\nDo you want to continue? [Y/n]"
```
genestack-solomatin: It is common practice to trigger capital letter by pressing enter. In our case it will be better to make them both in lowercase.

genestack-application-manager
```diff
@@ -438,6 +461,25 @@ def show_info(files, vendor_only, with_filename, no_filename):
...
+
+    choice = ''
+    while choice != 'Y' and choice != 'n':
```
genestack-solomatin: `while choice not in 'Yn'`:

genestack-application-manager
```diff
@@ -438,6 +461,25 @@ def show_info(files, vendor_only, with_filename, no_filename):
         first_file = False
 
 
+def prompt_removing_stable_version(application, apps_ids, version):
+    apps = []
```
genestack-solomatin: You can wrap function to `try/except KeyboardInterrupt` block and return `False`   
In that case you will print exit message if CTRL+C pressed

genestack-solomatin:
You need to add check like that:
```python
from genestack_client.utils import isatty
If not isatty() and not force:  
    raise GenestackException('... you cant call promt ... ')
```

genestack-solomatin:
Should we add check to `stable` command too?

PR #79, opened 2015-11-18
Yk/cla multiple files

Juhlchen:


genestack_client/cla.py
```diff
@@ -161,14 +161,16 @@ def replace_file_reference(self, accession, key, accession_to_remove, accession_
...
+        :param accession: file accession or accession list
         :param key: key for source files
         :param accession_to_remove: accession to remove
         :param accession_to_add: accession to add
         :return: None
         """
-        self.invoke('replaceFileReference', accession, key, accession_to_remove, accession_to_add)
+        self.invoke('replaceFileReference', self.__accession_list__(accession), key, accession_to_remove, accession_to_add)
 
+    def __accession_list__(self, accession):
```
genestack-solomatin: Double underscore is syntax for special methods in python. This mothod is private should be named as `__accession_list`

genestack_client/cla.py
```diff
@@ -118,24 +118,24 @@ def change_command_line_arguments(self, accession, params):
...
 
-        :param accession: file accession
+        :param accession: file accession or accession list
         :type accession: str
         :param params: list of commandlines to be set
         :type params: list
         :return: None
         """
-        self.invoke('changeCommandLineArguments', accession, params if isinstance(params, list) else [params])
+        self.invoke('changeCommandLineArguments', self.__accession_list__(accession), params if isinstance(params, list) else [params])
```
genestack-solomatin: `params if isinstance(params, list) else [params]` looks like do same thing to `params` as  `__accession_list__` to accessions

PR #84, opened 2015-12-07
remove EnumerationValues

alexander-polovtcev:


genestack_client/bio_metainfo.py
```diff
@@ -61,7 +61,7 @@ def add_organism(self, key, value):
         :type value: str
         :rtype: None
         """
-        self._add_value(key, value, 'Organism')
+        self.add_string(key, value)
```
vasiliy-bout: Maybe we should mark these methods as `@deprecated`? Because now we can use original `add_string` method...

PR #89, opened 2015-12-16
app-manager: wrap `sys.stdout` into Unicode-aware `StreamWriter`

klas-genestack:
Otherwise Python can't guess encoding when piping out text and fails on, say,
non-ASCII versions.

### DIFF
--- a/genestack-application-manager
+++ b/genestack-application-manager
@@ -9,6 +9,8 @@
 # actual or intended publication of such source code.
 #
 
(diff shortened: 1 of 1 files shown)


genestack-solomatin:
Please bump version to 0.3.1a1

klas-genestack:
Done already: https://github.com/genestack/python-client/blob/master/genestack_client/version.py

PR #101, opened 2016-02-29
Add missed methods to add metainfo values, remove tests

genestack-solomatin:


genestack_client/genestack_metainfo.py
```diff
@@ -147,6 +163,41 @@ def add_person(self, key, name, phone=None, email=None):
...
+        :param issue_date: journal issue date
+        :type issue_date: str
+        :param issue_number: journal issue number
+        :type issue_number: str
+        :param pages: pages in the journal issue
+        :type pages: str
+        :rtype: None
+        """
+        result = Metainfo._create_dict_with_type('publication')
+        result['identifiers'] = identifiers if identifiers is not None else {}
```
klas-genestack: Maybe test for `identifiers` emptiness? In case somebody adds empty list, for example.

genestack_client/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.3.1a2'
+__version__ = '0.3.1a3'
```
klas-genestack: I think we could safely bump version to *0.4a1* now

PR #117, opened 2016-03-28
Add message about debug to print of server exception

genestack-solomatin:


genestack_client/genestack_exceptions.py
```diff
@@ -53,8 +53,11 @@ def __str__(self):
                 self.message,
                 self.path
             )
-        if self.debug and self.stack_trace:
-            message += '\nStacktrace from server is:\n%s' % self.stack_trace
+        if self.stack_trace:
+            if self.debug:
+                message += '\nStacktrace from server is:\n%s' % self.stack_trace
+            else: message += '\nUse connection with "debug" option on to retrieve traceback'
```
klas-genestack: There's no point in _not_ breaking the line after `else:`

klas-genestack: *'Use connection with "debug" option on to retrieve traceback'* may confuse users, who use CL tool. Let's just say *'Enable debug option to retrieve traceback'*, I think that's enough in general case, and less confusing for those who don't know about `Connection`s.

genestack-solomatin:
updated

PR #118, opened 2016-03-28
Fix for `-r` and `-S system` at the same time

ekozobrodov:
[#5852](https://trac.genestack.com/ticket/5852)


genestack_client/scripts/genestack_application_manager.py
```diff
@@ -479,7 +479,11 @@ def upload_single_file(application, file_path, version, override,
         if result:
             print result
 
-        if release:
+        if release and stable and SCOPE_DICT[scope] == 'SYSTEM':
+            sys.stderr.write('WARNING: You are using `-r` flag with `-S system` construction at the same time. Setting '
+                             'scope `system` will automatically create released version, release it and set visible for'
+                             ' all. `-r` flag will be ignored.\n')
+        elif release:
             release_applications(application, app_info.identifiers, version, version + '-released', override)
```
genestack-solomatin: Move this code outside `try-except` and always check return status. Currently we ignore any errors occurred inside.

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -479,7 +479,11 @@ def upload_single_file(application, file_path, version, override,
         if result:
             print result
 
-        if release:
+        if release and stable and SCOPE_DICT[scope] == 'SYSTEM':
```
genestack-solomatin: In case then we have scope user and `-r` we will have next issue:

Application will be released with new version, and old version will be marked stable for user.

PR #122, opened 2016-04-04
Add new application

Blizarre:


genestack-solomatin:
Please bump version to `0.5.2a1`
https://github.com/genestack/python-client/blob/master/genestack_client/version.py

PR #123, opened 2016-04-05
View logs longer than 128k characters

daniel-ohayon:


daniel-ohayon:
Can be tested with the following:

```python
from genestack_client import get_connection, TaskLogViewer

# connect as Public on int-dev
connection = get_connection()
viewer = TaskLogViewer(connection)
viewer.view_long_log("GSF12945731")

**Comment by genestack-solomatin** [comment]
It is not user trouble if logs are long. User need to have only one function to get logs.

File: genestack_client/task_log_viewer.py [hunk 1]
```diff
@@ -31,75 +31,37 @@ def view_log(self, accession, log_type=None, follow=True):
...
-
         :param accession: file accession
         :param log_type: `stdout` or `stderr`
         :param follow: if enabled, wait and display new lines as they appear (similar to ``tail --follow``)
         """
         if not log_type:
             log_type = self.STDOUT
-        offset = -1
+        offset = 0
         limit = 128000
```
genestack-solomatin: Move this to keyword argument.

PR #124, opened 2016-04-11
Refactoring in shell

genestack-solomatin:
Return correct exit status for command.
  
  - `GenestackShell.process_command` now return its exit status
  - `Command.run` method return value is totally ignored, need to raise `GenestackException` to exit with status `1`
  - remove catch of `GenestackException` in application manager, this erros will be catched later and printed to stderr.

### DIFF
--- a/genestack_client/genestack_shell.py
+++ b/genestack_client/genestack_shell.py
@@ -15,7 +15,7 @@ import cmd
 import shlex
 from traceback import print_exc
 
-from genestack_client import GenestackAuthenticationException, GenestackServerException
+from genestack_client import GenestackAuthenticationException, GenestackException
 from version import __version__
 
 from utils import isatty, make_connection_parser, get_connection
@@ -144,6 +144,8 @@ class Command(object):
     def run(self):
         """
         Override this method to implement the command action.
+
+        :rtype: None
         """
         raise NotImplementedError()
 
@@ -258,8 +260,8 @@ class GenestackShell(cmd.Cmd):
                 parser = self.get_shell_parser(offline=True)
                 _, others = parser.parse_known_args()
 
-            self.process_command(command, others)
-            exit(0)
+            exit_code = self.process_command(command, others)
+            exit(exit_code)
 
         # do shell
         try:
@@ -298,7 +300,7 @@ class GenestackShell(cmd.Cmd):
 
     def process_command(self, command, argument_list, shell=False):
         """
-        Run a command with arguments.
+        Run a command with arguments, returns exit code.
 
         :param command: command
         :type command: Command
@@ -306,6 +308,8 @@ class GenestackShell(cmd.Cmd):
         :type argument_list: list
         :param shell: should we use shell mode?
         :type shell: bool
+        :return: exit code of command
+        :rtype: int
         """
         if shell or command.OFFLINE:
             p = command.get_command_parser()
@@ -314,21 +318,22 @@ class GenestackShell(cmd.Cmd):
         try:
             args = p.parse_args(argument_list)
         except SystemExit:
-            return
+            return 1
         command.set_connection(self.connection)
         command.set_arguments(args)
         try:
-            return command.run()
+            command.run()
+            return 0
         except (KeyboardInterrupt, EOFError):
             print
             print "Command interrupted."
-            return
-        except GenestackServerException as e:
+        except GenestackException as e:
             sys.stdout.flush()
             sys.stderr.write('%s\n' % e)
         except Exception:
             sys.stdout.flush()
             print_exc()
+        return 1
 
     def do_help(self, line):
         print
--- a/genestack_client/genestack_shell.py
+++ b/genestack_client/genestack_shell.py
@@ -309,7 +309,7 @@ class GenestackShell(cmd.Cmd):
 
     def process_command(self, command, argument_list, shell=False):
         """
-        Run a command with arguments, returns exit code.
+        Runs the given command with the provided arguments and returns the exit code
 
         :param command: command
         :type command: Command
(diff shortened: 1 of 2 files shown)


genestack_client/genestack_shell.py
```diff
@@ -298,14 +300,16 @@ def do_EOF(self, line):
 
     def process_command(self, command, argument_list, shell=False):
         """
-        Run a command with arguments.
+        Run a command with arguments, returns exit code.
```
alexander-polovtcev: I don't think this appendix is worth mentioning, it is obvious.
`Runs the given command with the provided arguments`

genestack_client/genestack_shell.py
```diff
@@ -298,14 +300,16 @@ def do_EOF(self, line):
...
 
         :param command: command
         :type command: Command
         :param argument_list: the list of arguments for the command
         :type argument_list: list
         :param shell: should we use shell mode?
         :type shell: bool
+        :return: exit code of command
```
alexander-polovtcev: `exit code of the given command`

alexander-polovtcev: actually this is not true, we don't return anything from the command itself

genestack_client/genestack_shell.py
```diff
@@ -314,21 +318,22 @@ def process_command(self, command, argument_list, shell=False):
         try:
             args = p.parse_args(argument_list)
         except SystemExit:
-            return
+            return 1
         command.set_connection(self.connection)
         command.set_arguments(args)
         try:
-            return command.run()
+            command.run()
```
alexander-polovtcev: why do we disable the possibility to return user-defined exit codes? And if the only option for the command to fail is throwing `GenestackException`, then we should at least mention it in the docstring of the `run` method

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -394,85 +393,62 @@ def resolve_jar_file(file_path):
...
-        for app_id in app_id_list:
-            sys.stdout.write('%-40s ... ' % app_id)
-            sys.stdout.flush()
-            application.invoke('markAsStable', app_id, scope, version)
-            sys.stdout.write('ok\n')
-            sys.stdout.flush()
-    except GenestackException as e:
-        sys.stderr.write('%s\n' % e.message)
-        return 1
+    print('Setting the application version "%s" stable for scope %s'
```
alexander-polovtcev: this can be one line

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -394,85 +393,62 @@ def resolve_jar_file(file_path):
...
+          % (version, scope))
+    scope = SCOPE_DICT[scope]
+    for app_id in app_id_list:
+        sys.stdout.write('%-40s ... ' % app_id)
```
alexander-polovtcev: Maybe we should disable `stdout` buferization in the scripts? It's quite tedious to write `sys.stdout.flush()` every time. See [this blog post](http://algorithmicallyrandom.blogspot.ru/2009/10/python-tips-and-tricks-flushing-stdout.html).

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -394,85 +393,62 @@ def resolve_jar_file(file_path):
...
         if get_system_stable_apps_version(application, app_info.identifiers, version):
-            sys.stderr.write('Can\'t install version "%s". This version is already system stable.\n' % version +
-                             'If you want to upload new version and make it stable, add "-S system" option.\n' +
-                             'Otherwise use another version name.\n')
-            return
-    if stable and release:
-        sys.stderr.write('Flags \'-r\' and \'-s\' cannot be used at once\n')
-        return
+            raise GenestackException('Can\'t install version "%s". This version is already system stable.\n' % version +
+                                     'If you want to upload new version and make it stable, add "-S system" option.\n' +
```
alexander-polovtcev: `If you want to upload a new version`

genestack_client/genestack_shell.py
genestack-solomatin: It is most important thing of command, It should be mentioned in docstring main part.

alexander-polovtcev: This description is still useless: what exit codes can it return? who returns this exit code, the command or the method? do these exit codes have any meanings?

genestack_client/genestack_shell.py
genestack-solomatin: We don't need them.   This make application is less complex.   Docstring to run added.

genestack_client/genestack_shell.py
```diff
@@ -145,6 +145,15 @@ def run(self):
         """
         Override this method to implement the command action.
 
+        Return value of this method is always ignored.
+        Exception thrown from this method means abnormal termination of the command.
```
alexander-polovtcev: `If this method raises an exception, the command will be treated as failed`

genestack_client/genestack_shell.py
```diff
@@ -144,6 +144,17 @@ def get_short_description(self):
     def run(self):
         """
         Override this method to implement the command action.
+
+        Return value of this method is always ignored.
+        Exception thrown from this method means abnormal termination of the command.
```
alexander-polovtcev: `If this method raises an exception, the command will be treated as failed`

genestack_client/genestack_shell.py
```diff
@@ -144,6 +144,17 @@ def get_short_description(self):
...
+
+        If this command is executed not in shell mode,
```
alexander-polovtcev: `If this command is executed in the shell mode, the failed state is ignored, otherwise exit code ``1`` is returned.

genestack_client/genestack_shell.py
```diff
@@ -298,14 +309,17 @@ def do_EOF(self, line):
...
-        Run a command with arguments.
+        Run a command with arguments, returns exit code.
 
         :param command: command
         :type command: Command
         :param argument_list: the list of arguments for the command
         :type argument_list: list
         :param shell: should we use shell mode?
         :type shell: bool
+        :return: exit status of command execution
```
alexander-polovtcev: `:return: 0 if the command was executed successfully, 1 otherwise`

PR #128, opened 2016-04-22
Added hack for backward compatibility

ekozobrodov:
And also:
- Incremented version;
- Added ChangeLog for last changes.

Main ticket: [#5879](https://trac.genestack.com/ticket/5879)
Ticket about removing hack for backward compatibility: [#5927](https://trac.genestack.com/ticket/5927)


genestack_client/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.5.2a1'
+__version__ = '0.5.2a2'
```
alexander-polovtcev: I think that the `version` should become `0.6.0`. Consult @genestack-solomatin about that.

alexander-polovtcev: also, there is an application, that checks the python-client version. We should bump it there too

PR #130, opened 2016-04-25
Don't expect latest version from server.

genestack-solomatin:

### DIFF
--- a/genestack_client/genestack_connection.py
+++ b/genestack_client/genestack_connection.py
@@ -109,12 +109,10 @@ class Connection:
 
         if compatible <= my_verison:
             return
-
-        update_message = ('You can update it with the following command:\n'
-                          '    pip install https://github.com/genestack/python-client/archive/stable.zip\n')
-
-        raise GenestackException('Your version "%s" is too old, please update to %s.\n%s' % (
-            my_verison, compatible, update_message))
+        raise GenestackException('Your Genestack Client version "%s" is too old, please update to %s.\n'
+                                 'You can update it with the following command:\n'
+                                 '    pip install https://github.com/genestack/python-client/archive/stable.zip' % (
+            my_verison, compatible))
 
     def logout(self):
         """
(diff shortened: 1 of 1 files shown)


genestack_client/genestack_connection.py
```diff
@@ -87,42 +87,34 @@ def login(self, email, password):
...
+
         update_message = ('You can update it with the following command:\n'
                           '    pip install https://github.com/genestack/python-client/archive/stable.zip\n')
 
-        if my_verison >= compatible:
-            return 'Newer version "%s" available, please update.\n%s' % (latest_version, update_message)
-        else:
-            raise GenestackException('Your version "%s" is too old, please update to %s.\n%s' % (
-                my_verison, latest_version, update_message))
+        raise GenestackException('Your version "%s" is too old, please update to %s.\n%s' % (
```
klas-genestack: 1. Let's drop `"to %s"`, because we advise to update to latest version anyway.

2. Let's elaborate on what version we're talking about: `"Your Genestack Client version"`, one can't be too clear.

3. Now it's weird to add `update_message`via string formatting, there should be only one string literal.

genestack-solomatin: 1. It is more clear if user can see which version he should update.
2. done
3. done

genestack_client/genestack_connection.py
```diff
@@ -109,12 +109,10 @@ def check_version(self, version):
 
         if compatible <= my_verison:
             return
-
-        update_message = ('You can update it with the following command:\n'
-                          '    pip install https://github.com/genestack/python-client/archive/stable.zip\n')
-
-        raise GenestackException('Your version "%s" is too old, please update to %s.\n%s' % (
-            my_verison, compatible, update_message))
+        raise GenestackException('Your Genestack Client version "%s" is too old, please update to %s.\n'
```
klas-genestack: > It is more clear if user can see which version he should update

If you insist let's make that:
```
Your Genestack Client version "%s" is too old, at least "%s" is required.
```

genestack-solomatin:
Done.

PR #137, opened 2016-05-20
Sa/exit with specified status on client version error 5987

genestack-solomatin:


genestack_client/genestack_connection.py
```diff
@@ -18,6 +18,7 @@
 from distutils.version import StrictVersion
 
 from genestack_client import GenestackServerException, GenestackAuthenticationException, GenestackException, __version__
+from genestack_client.genestack_exceptions import GenestackVersionException
```
klas-genestack: Why are exceptions imported from two different modules?

genestack-solomatin: Fixed

klas-genestack: Really?  I can't see fix in `genestack_connection`

PR #146, opened 2016-06-23
License

genestack-solomatin:


vasiliy-bout:
We must install LICENCE.txt to clients

PR #149, opened 2016-06-30
Encode exception message

genestack-solomatin:
Fix crash on exception print.

Steps to reproduce:
  - start genestack-shell
  - type `ы`

### DIFF
--- a/genestack_client/genestack_exceptions.py
+++ b/genestack_client/genestack_exceptions.py
@@ -26,11 +26,7 @@ class GenestackServerException(GenestackException):
         :param stack_trace: server stack trace
         :type stack_trace: str
         """
-        def to_str(val):
-            if isinstance(val, unicode):
-                return val.encode('utf-8', 'ignore')
-            return val
-        message = to_str(message)
+        message = message.encode('utf-8', 'ignore') if isinstance(message, unicode) else message
 
         GenestackException.__init__(self, message, path, post_data, debug, stack_trace)
         self.message = message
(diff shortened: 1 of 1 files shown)


genestack_client/genestack_exceptions.py
```diff
@@ -26,6 +26,12 @@ def __init__(self, message, path, post_data, debug=False, stack_trace=None):
...
         :type stack_trace: str
         """
+        def to_str(val):
+            if isinstance(val, unicode):
+                return val.encode('utf-8', 'ignore')
+            return val
+        message = to_str(message)
+
         GenestackException.__init__(self, message, path, post_data, debug, stack_trace)
         self.message = message
```
alexander-polovtcev: why do you store the message twice - in the superclass and here?

genestack-solomatin: In superclass this stored in `args` attribute.  Passing arguments to superclass required for pickling exception if it is used in threads. http://stackoverflow.com/a/28335286

genestack_client/genestack_exceptions.py
```diff
@@ -26,6 +26,12 @@ def __init__(self, message, path, post_data, debug=False, stack_trace=None):
         :param stack_trace: server stack trace
         :type stack_trace: str
         """
+        def to_str(val):
+            if isinstance(val, unicode):
```
alexander-polovtcev: use ternary operator here

PR #155, opened 2016-07-19
Add keyring dependencies

genestack-solomatin:
There is trouble on installing python client on ubuntu.  

Keyring 8.0 have changes that don't see default security backend on ubunty.  https://github.com/jaraco/keyring/issues/117

Simple solution is to install `keyrings.alt` package, but It is possible that FileBackend will be used(base64 encoded pasword in plain text file).

To test your platform you need to install `keyring` and `keyrings.alt` and execute code:

``` python
import keyring.core
print keyring.core._keyring_backend
```


klas-genestack:
> Simple solution is to install `keyrings.alt` package, but It is possible that FileBackend will be used(base64 encoded pasword in plain text file).

In which cases this would be used?  We could check if it's selected backend and fail/warn.

setup.py
```diff
@@ -13,7 +13,7 @@
     author='Genestack Limited',
     author_email='',
     description='Genestack Python Client Library',
-    install_requires=['keyring', 'requests'],
+    install_requires=['keyring', 'requests', 'keyrings.alt'],
```
klas-genestack: How about
```python
install_requires=['keyring', 'requests'] + ['keyrings.alt'] if sys.platform.startswith('linux') else [],
```

genestack-solomatin:
On ubuntu
```
>>> import keyring.core
>>> print keyring.core._keyring_backend
<keyring.backends.file.EncryptedKeyring object at 0x7f8d24c44750>
```

klas-genestack:
Yeah, I know, but it doesn't answer my question.
You say that insecure backend could be used, so I ask "when" and "should we guard against it".

klas-genestack:
[My question](https://github.com/genestack/python-client/pull/155#issuecomment-233904989) still stands.

genestack-solomatin:
Not actual.

PR #160, opened 2016-08-03
docs: use absolute path to import project, otherwise it doesn't work

klas-genestack:


genestack-solomatin:
Need to check this for task-library!

PR #164, opened 2016-08-09
Enable debug from within app-manager shell. fixes #5988

genestack-solomatin:


genestack_client/genestack_shell.py
```diff
@@ -291,7 +291,7 @@ def set_shell_user(self, args):
             self.prompt = '%s> ' % email
         except GenestackAuthenticationException:
             self.prompt = 'anonymous>'
-        self.intro = '\ngenestack_client v%s\n%s' % (__version__, self.INTRO)
+        self.intro = 'genestack_client v%s\ndebug: %s\n%s' % (__version__, self.connection.debug, self.INTRO)
```
klas-genestack: It's really hard to read (`"%s\n"`), could we use a multiline template string?

klas-genestack: Also printing "debug: False" seems excessive.

genestack_client/genestack_shell.py
```diff
@@ -342,14 +342,18 @@ def process_command(self, command, argument_list, shell=False):
             print_exc()
         return 1
 
+    def do_debug(self, line):
```
klas-genestack: `toggle_debug`, maybe?

PR #167, opened 2016-08-09
Apply openSSL only for linux

genestack-solomatin:
It is trouble on building pyOpenSSL on latest maxOSx (El capitan)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,9 @@
 # -*- coding: utf-8 -*-
 
 from distutils.core import setup
+
+import sys
+
 exec(open('genestack_client/version.py').read())
 
 setup(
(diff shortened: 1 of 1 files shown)


Blizarre:
hi, what is pyopenssl used for in the python client library ?

setup.py
```diff
@@ -13,7 +16,7 @@
     author='Genestack Limited',
     author_email='',
     description='Genestack Python Client Library',
-    install_requires=['keyring', 'requests', 'pyOpenSSL'],
+    install_requires=['keyring', 'requests'] + ['pyOpenSSL'] if sys.platform.startswith('linux') else [],
```
Blizarre: what will happen there then :

https://github.com/genestack/python-client/blob/65208f4630e6eff2e08e0d47e073a4d9c066a398/genestack_client/chunked_upload.py#L13

genestack-solomatin: Thank you, I missed that I add this dependency.

PR #169, opened 2016-08-12
Remove old dictionary support

genestack-solomatin:


ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 0.9.1
```
klas-genestack: It's bad to deprecate or drop deprecated stuff in minor versions.
Let's leave it for 0.10.

PR #172, opened 2016-08-24
Add term type to dictionary uploader

genestack-solomatin:


ChangeLog
```diff
@@ -1,5 +1,6 @@
 * 0.10.0
-  * remove `DictionaryUtil`
```
klas-genestack: "Removed" and "Added"
http://keepachangelog.com/en/0.3.0/ (also http://keepachangelog.com/ru/0.3.0/)

genestack_client/data_importer.py
```diff
@@ -575,6 +575,9 @@ def create_dictionary(self, parent=None, name=None, url=None, metainfo=None):
         :param name: name of the file
         :type name: str
         :param url: URL of a file
+        :type url: str
```
klas-genestack: It's a bit misplaced, but all right

genestack_client/data_importer.py
```diff
@@ -583,6 +586,7 @@ def create_dictionary(self, parent=None, name=None, url=None, metainfo=None):
         metainfo = metainfo or BioMetainfo()
         name and metainfo.add_string(BioMetainfo.NAME, name)
         url and metainfo.add_external_link(BioMetainfo.DATA_LINK, url)
+        term_type and metainfo.add_string('genestack.dictionary:termType', term_type)
```
klas-genestack: I strongly dislike using `and` in this way instead of `if` (it should be used in assignments only), but since it's all over the `DataImporter`, I'll just grumble a bit.

PR #176, opened 2016-08-24
Add check for arguments

genestack-solomatin:


genestack_client/data_importer.py
```diff
@@ -122,6 +122,19 @@ def __invoke_loader(self, parent, importer_type, metainfo):
         self.__process_links(metainfo)
         return self.importer.invoke('createFile', parent, importer_type, metainfo)['accession']
 
+    def __add_to_metainfo(self, metainfo, key, value, setter, required=False):
+        if value is None:
+            if required:
```
klas-genestack: `if required and metainfo.get(key) is None:`?

genestack_client/data_importer.py
```diff
@@ -122,6 +122,19 @@ def __invoke_loader(self, parent, importer_type, metainfo):
         self.__process_links(metainfo)
         return self.importer.invoke('createFile', parent, importer_type, metainfo)['accession']
 
+    def __add_to_metainfo(self, metainfo, key, value, setter, required=False):
```
klas-genestack: I've written a couple of lines before realizing that I was wrong about some details behind the function =)
So, there's a need for doctsring!

genestack_client/data_importer.py
```diff
@@ -122,6 +122,19 @@ def __invoke_loader(self, parent, importer_type, metainfo):
...
+        if value is None:
+            if required:
+                if metainfo.get(key) is None:
+                    raise GenestackException('Missed required key "%s", '
+                                             'it should be specified as function argument '
+                                             'or in metainfo' % key)
```
klas-genestack: `'or in metainfo object'`

genestack_client/data_importer.py
```diff
@@ -122,6 +122,19 @@ def __invoke_loader(self, parent, importer_type, metainfo):
...
+        else:
+            if metainfo.get(key) is not None:
+                raise GenestackException('Key "%s", is passed both as function argument '
```
klas-genestack: Bad comma (`,`), remove it.

klas-genestack:
The idea seems solid, but I'd rewrite the functions thus:
```py
    def __add_to_metainfo(self, metainfo, key, value, setter, required=False):
        mi_key = metainfo.get(key)
        if required and value is None and mi_key is None:
            raise GenestackException('Missing required key "%s", '
                                     'it should be passed as function argument '
                                     'or in metainfo object' % key)
        if value is not None and mi_key is not None:
            raise GenestackException('Key "%s" is passed both as function argument '
                                     'and inside metainfo object' % key)
        value = value if value is not None else mi_key
        setter(key, value)
```

genestack_client/data_importer.py
```diff
@@ -124,6 +124,35 @@ def __invoke_loader(self, parent, importer_type, metainfo):
         self.__process_links(metainfo)
         return self.importer.invoke('createFile', parent, importer_type, metainfo)['accession']
 
+    def __add_to_metainfo(self, metainfo, key, value, setter, required=False):
+        """
+        Add value to metainfo, check if value is not already present,
+         checks required fields.
```
genestack-solomatin: remove space

genestack_client/data_importer.py
genestack-solomatin: done

genestack_client/data_importer.py
```diff
@@ -182,9 +214,11 @@ def create_vcf(self, parent=None, name=None, reference_genome=None, url=None, me
         :rtype: str
         """
         metainfo = metainfo or Metainfo()
-        name and metainfo.add_string(Metainfo.NAME, name)
+        self.__add_to_metainfo(metainfo, Metainfo.NAME, name,
+                               metainfo.add_string, required=True)
         reference_genome and metainfo.add_file_reference(BioMetaKeys.REFERENCE_GENOME, reference_genome)
```
genestack-solomatin: missed

genestack-solomatin:
> The idea seems solid, but I'd rewrite the functions thus:

No, we should add value only if it is missed in metainfo

genestack_client/data_importer.py
```diff
@@ -336,12 +375,14 @@ def create_sequencing_assay(self, parent, name=None, urls=None,
...
-        method and metainfo.add_string(BioMetaKeys.METHOD, method)
+        self.__add_to_metainfo(metainfo, Metainfo.NAME, name, StringValue, required=True)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.ORGANISM, organism, StringValue)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.METHOD, method, StringValue)
         if urls:
+            urls = (url for url in urls)
+            self.__add_to_metainfo(metainfo, BioMetaKeys.READS_LINK, urls.next(), ExternalLink)
             for url in urls:
-                metainfo.add_external_link(BioMetaKeys.READS_LINK, url)
+                metainfo.add_value(BioMetaKeys.READS_LINK, ExternalLink(url))
```
klas-genestack: Do we intend to deprecate `add_internal_link()`?

genestack_client/data_importer.py
```diff
@@ -490,15 +532,19 @@ def create_reference_genome(self,
...
-            metainfo.add_external_link(SEQUENCE_KEY, seq_link, text='Sequence data link')
+        self.__add_to_metainfo(metainfo, Metainfo.NAME, name, StringValue, required=True)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.ORGANISM, organism, StringValue)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.STRAIN, strain, StringValue)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.REFERENCE_GENOME_ASSEMBLY, assembly, StringValue)
+        self.__add_to_metainfo(metainfo, BioMetaKeys.REFERENCE_GENOME_RELEASE, release, StringValue)
+        self.__add_to_metainfo(metainfo, ANNOTATION_KEY, annotation_url, ExternalLink)
+        self.__add_to_metainfo(metainfo, metainfo.DESCRIPTION, description, StringValue)
+
+        if sequence_urls:
```
klas-genestack: Why is it so complicated?  Old code was much clearer and more straightforward.

klas-genestack: Why is it so complicated? Old code was much clearer and more straightforward.

genestack_client/data_importer.py
genestack-solomatin: We just added clases of metainfo values, I did not think about this yet.

genestack_client/data_importer.py
genestack-solomatin: `add_value` accepts only scalar value.
We want to pass list of values and make check for first item. 

It think about changing `add_value` to accept list as value, but not in this PR.

genestack_client/data_importer.py
klas-genestack: So, what's the purpose of this change (and similar ones)?  `add_external_link` calls `add_value` under the hood anyway, so I see no reason to make these changes.

genestack_client/data_importer.py
klas-genestack: I dislike this intensely.
1. Code looks ugly and way too complicated for the simple task.
2. I can't see why we test only the first URL.  And test so obliquely at that!  It's all urls in list, let's treat them all equally.

genestack-solomatin: > I can't see why we test only the first URL

Because if test fro first url is passed, we don't need to test other and cannot test other because this will provide an error.

I can move this test into  test function (if is_list ....) or wait a little and implement ListValue support

klas-genestack: > we don't need to test other and cannot test other because this will provide an error

Why?

genestack-solomatin: Add first with check

ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 1.13.0
```
klas-genestack: 1.13.0?

klas-genestack:
What's the status of this?

genestack_client/data_importer.py
```diff
@@ -124,6 +125,40 @@ def __invoke_loader(self, parent, importer_type, metainfo):
         self.__process_links(metainfo)
         return self.importer.invoke('createFile', parent, importer_type, metainfo)['accession']
 
+    @staticmethod
+    def __add_to_metainfo(metainfo, key, value, value_type, required=False):
+        """
+        Add value to metainfo, check if value is not already present,
+        checks required fields.
```
klas-genestack:
```rst
Add ``key``: ``value`` pair to ``metainfo``.

Fails if ``metainfo`` already contains ``key`` or ``value`` is ``required``, but empty.
```

genestack_client/data_importer.py
```diff
@@ -124,6 +125,40 @@ def __invoke_loader(self, parent, importer_type, metainfo):
...
+        """
+        current_value = metainfo.get(key)
+        if current_value is not None and value is not None:
+            raise GenestackException('Key "%s" is passed both as function argument '
+                                     'and inside metainfo object' % key)
+        if current_value:
+            return
+
+        if required and value is None:
+                raise GenestackException('Missing required key "%s", '
```
klas-genestack: Bad indentation.

genestack_client/data_importer.py
```diff
@@ -124,6 +125,40 @@ def __invoke_loader(self, parent, importer_type, metainfo):
...
+                                         'it should be passed as function argument '
+                                         'or in metainfo object' % key)
+        if value is not None:
```
klas-genestack: But when `required == False` and `value is None` we don't add `key: None` to metainfo.
Is it expected behaviour?  Feels wrong to me, but if it is, it warrants a comment.

genestack_client/data_importer.py
klas-genestack: Add ``key``: ``value`` pair to ``metainfo``.

Fails if:
* ``metainfo`` already contains ``key``;
* ``key`` is ``required``, but no value is provided by user,
  and neither ``metainfo`` has already existing ``key`` (with some value).

PR #178, opened 2016-08-28
Deprecate BioMetainfo

genestack-solomatin:


genestack_client/bio_metainfo.py
```diff
@@ -1,48 +1,20 @@
...
-    DNASEQ_TECHNOLOGY = 'Genome variation profiling by high throughput sequencing'
-    CHIPSEQ_TECHNOLOGY = 'Genome binding/occupancy profiling by high throughput sequencing'
-    MICROARRAY_TECHNOLOGY = 'Expression profiling by array'
-    DNA_MICROARRAY_TECHNOLOGY = 'Genome variation profiling by SNP array'
-    # SAMPLE_ID = 'genestack.bio:sampleId'      # still used in 1000 loader.
-    EXTRACTED_MOLECULE = 'genestack.bio:extractedMolecule'
-    PLATFORM = 'genestack.bio:platform'
-    SOURCE_DATA_PREFIX = 'sourceData:'
-    SECONDARY_ACCESSION = 'genestack.bio:secondaryAccession'
+    def __init__(self, iterable=None, **kwargs):
```
klas-genestack: If user imports `BioMetainfo`, but uses it only as a collection of constants, never initiating instances, then (s)he won't get any deprecation message.  The way to implement more correct behaviour seems to make `BioMetainfo` an instance of metaclass with overriden `__getattribute__()`.  Instead of multiple inheritance, just return attributes of `BioMetaKeys` (and complain about using deprecated class, of course).

PR #181, opened 2016-09-20
Batched container children retrieval

daniel-ohayon:
Related to [this pull request](https://github.com/genestack/genestack-applications/pull/68)


genestack_client/files_util.py
```diff
@@ -120,16 +121,41 @@ def collect_initializable_files_in_container(self, accession):
...
+        :type show_progress: bool
         :return: list of accessions
         :rtype: list
         """
-        return self.invoke('getFileChildren', container_accession)
+        all_files = []
+        count = 0
+        while True:
+            if show_progress:
+                sys.stdout.write('\rRetrieving container children (%d)...' % count)
```
genestack-solomatin: Should it start with line break in TTY mode?

daniel-ohayon: Depends who you ask. In sublime text it looks good because ^R is rendered as a newline character, but in `less`, I see that:
```
^MRetrieving container children (0)...^MRetrieving container children (500)...^M
```

PR #187, opened 2016-10-27
Prompt user to update if API version check fails

beneills:
If there's a serious server error, we display:

"We encountered an unknown error while querying the Genestack API.  Please verify that you're running the latest version via: `pip install https://github.com/genestack/python-client/archive/stable.zip --upgrade`"

### DIFF
--- a/genestack_client/genestack_connection.py
+++ b/genestack_client/genestack_connection.py
@@ -95,16 +95,16 @@ class Connection:
         :param version: version in format suitable for distutils.version.StrictVersion
         :return: None
         """
+        my_version = StrictVersion(__version__)
+
         try:
             compatible_version = self.application('genestack/clientVersion'
                                                  ).invoke('getCompatibleVersion')
         except GenestackServerException, e:
             # We don't know what happened, but it might be due to incompatible client/API versions.
-            # Rethrow exception, making sure we tell the user to update.
-            e.prompt_update = True
-            raise e
+            # Throw a version exception, making sure we tell the user to update.
+            raise GenestackVersionException(my_version)
 
-        my_version = StrictVersion(__version__)
         compatible = StrictVersion(compatible_version)
 
         if compatible <= my_version:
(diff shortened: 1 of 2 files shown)


genestack_client/genestack_exceptions.py
```diff
@@ -28,12 +28,13 @@ def __init__(self, message, path, post_data, debug=False, stack_trace=None):
...
         message = message.encode('utf-8', 'ignore') if isinstance(message, unicode) else message
 
-        GenestackException.__init__(self, message, path, post_data, debug, stack_trace)
+        GenestackException.__init__(self, message, path, post_data, debug, stack_trace, prompt_update)
         self.message = message
         self.debug = debug
         self.stack_trace = stack_trace
         self.path = path
         self.post_data = post_data
+        self.prompt_update = prompt_update
```
vasiliy-bout: This flag looks like a dirty hack. We must not merge this branch.

genestack_client/genestack_connection.py
```diff
@@ -95,8 +95,14 @@ def check_version(self):
...
-        compatible_version = self.application('genestack/clientVersion'
-                                             ).invoke('getCompatibleVersion')
+        try:
+            compatible_version = self.application('genestack/clientVersion'
+                                                 ).invoke('getCompatibleVersion')
+        except GenestackServerException, e:
+            # We don't know what happened, but it might be due to incompatible client/API versions.
+            # Rethrow exception, making sure we tell the user to update.
+            e.prompt_update = True
+            raise e
```
vasiliy-bout: We can simply throw here a `GenestackVersionException` with corresponding message.

Currently message is hardcoded inside `GenestackVersionException`. We can change this logic and provide the message to constructor.

genestack-solomatin:
And there is one more idea, replace exception throw with.
```
sys.stderr.write(message)
exit(13)  
```

genestack_client/genestack_connection.py
beneills: Fixed.

genestack_client/genestack_exceptions.py
beneills: Fixed.

PR #190, opened 2016-10-31
Update error message on version check

genestack-solomatin:
Show latest compatible version for user at version update error.

### DIFF
--- a/genestack_client/genestack_exceptions.py
+++ b/genestack_client/genestack_exceptions.py
@@ -70,15 +70,22 @@ class GenestackVersionException(GenestackException):
     """
 
     def __init__(self, my_version, compatible=None):
+        """
+
+        :param my_version: current version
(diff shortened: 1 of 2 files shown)


genestack_client/genestack_exceptions.py
```diff
@@ -70,15 +70,22 @@ class GenestackVersionException(GenestackException):
...
+
+        :param my_version: current version
+        :type my_version: distutils.version.StrictVersion
+        :param compatible: compatible version
+        :type compatible: distutils.version.StrictVersion
+        """
         if compatible:
             required_message = ', at least "{req_version}" is required.'.format(req_version=compatible)
-            package = 'v{req_version}.zip'.format(req_version=compatible)
+            package = 'master' if compatible.prerelease else 'stable'
```
klas-genestack: Let's call it "branch" to be clear.

PR #191, opened 2016-11-08
Improve error message if cannot get required version

genestack-solomatin:
Old message was confusing

### DIFF
--- a/genestack_client/genestack_exceptions.py
+++ b/genestack_client/genestack_exceptions.py
@@ -69,26 +69,26 @@ class GenestackVersionException(GenestackException):
     Exception thrown if old version of client is used.
     """
 
-    def __init__(self, my_version, compatible=None):
+    def __init__(self, current_version, required_version=None):
         """
 
-        :param my_version: current version
-        :type my_version: distutils.version.StrictVersion
-        :param compatible: compatible version
-        :type compatible: distutils.version.StrictVersion
+        :param current_version: current version
+        :type current_version: distutils.version.StrictVersion
+        :param required_version: minimum required version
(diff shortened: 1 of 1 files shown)


genestack_client/genestack_exceptions.py
```diff
@@ -78,14 +78,17 @@ def __init__(self, my_version, compatible=None):
         :type compatible: distutils.version.StrictVersion
         """
         if compatible:
```
klas-genestack: As I understand now, this is an unfortunate argument name.
Should be `current_version` and `required_version`.

Moreover, it's weird to have to pass current version of Python Client to throw this exception.

genestack_client/genestack_exceptions.py
```diff
@@ -78,14 +78,17 @@ def __init__(self, my_version, compatible=None):
...
-            required_message = ', at least "{req_version}" is required.'.format(req_version=compatible)
             branch = 'master' if compatible.prerelease else 'stable'
+            message = (
+                'Your Genestack Client version "{version}" is too old, '
+                'at least "{req_version}" is required.\n'
+                       ).format(version=my_version, req_version=compatible)
         else:
-            required_message = ''
             branch = 'stable'
+            message = 'Cannot get required version from server.\n'.format(version=my_version)
```
klas-genestack:
```py
>>> 'Cannot get required version from server.'.format(version="WHERE EXACTLY IS {version} IN THIS TEMPLATE???!?!?!?!!!!1111")
'Cannot get required version from server.'
```

klas-genestack:
Let's bump minor version, then

PR #194, opened 2017-01-13
Share only folder without children

kheleon:
according to Daniel’s comment


ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 0.14.0a1
+  * Added ability to share specified folder only without its children (when `recursive` is set to false in ShareUtils)
```
genestack-solomatin: remove `only` or add coma after it.

PR #196, opened 2017-02-28
Add check if no files was found. Fixes #7332

genestack-solomatin:
`match_jar_globs` will return empty list if folder is not found


genestack_client/scripts/genestack_application_manager.py
```diff
@@ -405,7 +407,7 @@ def resolve_jar_file(file_path):
         raise GenestackException('More than one JAR file was found inside %s:\n'
                                  ' %s' % (file_path, '\n '.join(jar_files)))
     elif not jar_files:
-        raise GenestackException('No JAR file was found inside %s' % file_path)
+        raise GenestackException('No JAR file was found inside: "%s"' % file_path)
```
klas-genestack: "No JAR files were found within given files/directories: %s"

PR #197, opened 2017-03-06
Ek/5978 add status command

ekozobrodov:
Added `status` command for getting information about application loading, for more details see: [#5978](https://trac.genestack.com/ticket/5978)

Command output now looks like this:

```
genestack/qiime-report   failed

Application jar was not loaded due to the following errors:
genestack/qiime-report JS error: this language feature is only supported for ECMASCRIPT6 mode or better: class. Use --language_in=ECMASCRIPT6 or ECMASCRIPT6_STRICT or higher to enable ES6 features.
Script loading error: unable to find script dose_response.r
genestack/qiime-report JS error: this language feature is only supported for ECMASCRIPT6 mode or better: member declaration. Use --language_in=ECMASCRIPT6 or ECMASCRIPT6_STRICT or higher to enable ES6 features.


genestack/report-viewer   loaded
```

Also changed interface of `install` command.

Note: don't merge this branch until main repo branch will be finished (consult me on any questions).

Pull request in main repo: https://github.com/genestack/genestack/pull/3269 (based on [3253](https://github.com/genestack/genestack/pull/3253))


genestack_client/scripts/genestack_application_manager.py
```diff
@@ -103,8 +103,8 @@ def update_parser(self, p):
                  '`-i <group_accession>` for group visibility)'
         )
         p.add_argument(
-            '-w', '--wait', action='store_true',
-            help='wait until all installed applications will be completely loaded'
+            '-n', '--no-wait', action='store_true', dest='no_wait',
+            help='Don\'t wait until all installed applications will be completely loaded'
```
genestack-solomatin: We use `cannot` in such cases.

PS. Double quotes should be used to avoid escaping.  https://trac.genestack.com/wiki/DevStuff/CodeStyle/Python

genestack_client/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.14.0'
+__version__ = '0.15.0a'
```
genestack-solomatin: 0.15.0a1

ekozobrodov:
About `Use --language_in=ECMASCRIPT6`: i looked to closure compiler sources and it seems that there is no way to change error messages. It uses `ErrorReporter` and there are several implementations, but only one can report meaningful error messages.

genestack-solomatin:
Please bump version to '0.15.0a2'

PR #198, opened 2017-03-17
Change invoke params 7475

genestack-solomatin:
This PR should be merged after https://github.com/genestack/genestack/pull/3300


genestack_client/genestack_connection.py
```diff
@@ -221,7 +220,10 @@ def __init__(self, connection, application_id=None):
         if len(self.application_id.split('/')) != 2:
             raise GenestackException('Invalid application ID, expect "{vendor}/{application}" got: %s' % self.application_id)
 
-    def __invoke(self, path, post_data):
+    def __invoke(self, path, post_data, trace=None):
+        self.connection.opener.addheaders = [('gs-extendSession', 'true')]
+        if trace:
+            self.connection.opener.addheaders.append(('gs-trace', 'true'))
         f = self.connection.open(path, post_data)
```
vasiliy-bout: Fix `connection.open` docstring, error is here: `:rtype: urllib.addinfourl`

genestack_client/genestack_connection.py
```diff
@@ -221,7 +220,10 @@ def __init__(self, connection, application_id=None):
         if len(self.application_id.split('/')) != 2:
             raise GenestackException('Invalid application ID, expect "{vendor}/{application}" got: %s' % self.application_id)
 
-    def __invoke(self, path, post_data):
+    def __invoke(self, path, post_data, trace=None):
+        self.connection.opener.addheaders = [('gs-extendSession', 'true')]
+        if trace:
+            self.connection.opener.addheaders.append(('gs-trace', 'true'))
```
vasiliy-bout: We can rewrite it using `BaseHandler.protocol_open` method (see here: https://docs.python.org/2/library/urllib2.html#basehandler-objects).

But this seems like an overkill, so I propose just to move this logic into `connection.open` method, and send there just a multimap or another headers object implementation containing desired headers, so that the opener will be a private field (can we change the field declaration to be private?)

genestack_client/genestack_connection.py
```diff
@@ -253,16 +255,14 @@ def get_response(self, method, params=(), trace=True):
...
         """
-        post_data = {'method': method}
-        if trace:
-            post_data['trace'] = True
+        post_data = {}
         if params:
             post_data['parameters'] = json.dumps(params)
 
-        path = '/application/invoke/%s' % self.application_id
+        path = '/application/invoke/%s/%s' % (self.application_id, urllib.quote(method))
```
vasiliy-bout: Should we check that method is not unicode, and otherwise convert it with `encode('utf-8')`?

genestack_client/genestack_connection.py
genestack-solomatin: It is not an error, this class is returned. It is file like object with info(headers) and  url.

genestack_client/genestack_connection.py
```diff
@@ -151,20 +150,26 @@ def logout(self):
...
+        :type headers: list[tuple[str]]
         :return: response
         :rtype: urllib.addinfourl
         """
         if data is None:
             data = ''
         elif isinstance(data, dict):
             data = urllib.urlencode(data)
+        headers = list(headers)
+        headers.append(('gs-extendSession', 'true'))
```
vasiliy-bout: Invalid header name. It was renamed in the core repo.

genestack_client/genestack_connection.py
```diff
@@ -151,20 +150,26 @@ def logout(self):
...
+        :param headers: additional headers as list of pairs
+        :type headers: list[tuple[str]]
         :return: response
         :rtype: urllib.addinfourl
         """
         if data is None:
             data = ''
         elif isinstance(data, dict):
             data = urllib.urlencode(data)
+        headers = list(headers)
```
vasiliy-bout: Possible error here:
```
>>> list(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

genestack_client/genestack_connection.py
genestack-solomatin: This header was not touched.  

We have 3 headers in total, but they never present all three in same file :)

`gs-extendSession`, used in `python` and `js`,  handled outside `InvokeController`.
`Genestack-ExternalCall`  used in `js` and `InvokeController`
Genestack-Trace used in `python` and `InvokeController`

PR #199, opened 2017-03-22
Docstring update

genestack-solomatin:
- fix wrong rtype
 - add missed rtype
 - cosmetic changes


genestack_client/utils.py
```diff
@@ -93,6 +93,7 @@ def ask_confirmation(question, default=None):
     :param question: question to ask, without [y/n] suffix and question mark.
     :param default: default value for empty string. Can be ``'y'``, ``'n'``, and ``None``
     :return:
```
beneills:
```python
:return: whether user confirms
```

PR #203, opened 2017-04-14
DataImporter: don't mutate argument

genestack-solomatin:


genestack_client/data_importer.py
```diff
@@ -155,7 +155,7 @@ def create_bed(self, parent=None, name=None, reference_genome=None, url=None, me
         :return: file accession
         :rtype: str
         """
-        metainfo = metainfo or Metainfo()
+        metainfo = Metainfo(metainfo) if metainfo else Metainfo()
```
alexander-polovtcev: I think this should be a function

alexander-polovtcev: why didn't you replace all occurrences?

genestack_client/data_importer.py
```diff
@@ -247,14 +247,17 @@ def create_bam(self,
...
+        metainfo = self._get_metainfo(metainfo)
         name and metainfo.add_string(Metainfo.NAME, name)
         organism and metainfo.add_string(BioMetaKeys.ORGANISM, organism)
         strain and metainfo.add_string(BioMetaKeys.STRAIN, strain)
         reference_genome and metainfo.add_file_reference(BioMetaKeys.REFERENCE_GENOME, reference_genome)
         url and metainfo.add_external_link(BioMetaKeys.BAM_FILE_LINK, url)
         return self.__invoke_loader(parent, 'alignedReads', metainfo)
 
+    def _get_metainfo(self, metainfo):
+        return Metainfo(metainfo) if metainfo else Metainfo()
```
alexander-polovtcev: looks like this should be a static method and named like `copy_metainfo`

PR #206, opened 2017-04-25
python-client: add FEATURE_LIST constant to FilesUtil and DataImporter.create_gene_list

Michael31416:
This PR adds constant `FEATURE_LIST` and method `DataImporter.create_gene_list`, which is for importing so called gene lists. Each gene list is really a kind of feature list with some additional restrictions.


genestack-solomatin:
Bump version in `version.py`, current released is `0.150`, so you need to add at least `0.16.0a1`. It must be greater than version in `GenestackClientVersion`

genestack_client/data_importer.py
```diff
@@ -590,6 +590,37 @@ def create_mapped_reads_count(self,
         url and metainfo.add_external_link(BioMetaKeys.DATA_LINK, url)
         return self.__invoke_loader(parent, 'mappedReadCounts', metainfo)
 
+    def create_gene_list(self,
+                                  parent=None,
+                                  name=None,
+                                  url=None,
+                                  organism=None,
+                                  metainfo=None):
```
klas-genestack: This indenting style is too weird even for us.
Please format signature like `create_sequencing_assay()` (preferred) or at least like `create_bam()`.

Michael31416: @klas-genestack I've added a new commit with fixes

PR #209, opened 2017-05-04
#7667 Added ability to view module loading warnings

ekozobrodov:
See [#7667](https://trac.genestack.com/ticket/7667)

Appropriate pull request in main repo: https://github.com/genestack/genestack/pull/3415


ChangeLog
```diff
@@ -1,3 +1,7 @@
+* 0.15.1
+    * Added support of 'warnings' to `status` command
```
vasiliy-bout: What is the style in single quotes?

vasiliy-bout: And by the way, maybe it's time to rename this file into one with proper extension? `ChangeLog.md`?

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -536,6 +533,15 @@ def upload_single_file(application, file_path, version, override,
     if not no_wait:
```
vasiliy-bout: We added a message if we are waiting, but we should also add a message if we are not waiting.

If we are not waiting, we should remind users that if JAR was successfully loaded then there could be other errors and warnings that must be checked using `status` command manually by user.

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -536,6 +533,15 @@ def upload_single_file(application, file_path, version, override,
...
         for app_id in app_info.identifiers:
             wait_application_loading(application, app_id, version)
+        if app_info.identifiers:
+            # Get jar loading messages from one of its applications:
+            descriptor = application.invoke('getApplicationDescriptor', app_info.identifiers[0], version)
```
vasiliy-bout: Reuse `get_application_descriptor`

vasiliy-bout: We can reuse descriptor from `wait_application_loading`. Try to change method `wait_application_loading` into `do_when_application_loaded` that takes handler function as a parameter like in JS.

genestack_client/scripts/genestack_application_manager.py
```diff
@@ -605,20 +611,37 @@ def wait_application_loading(application, app_id, version, seconds=1):
...
+            return False
     return True
 
 
+def format_loading_messages_by_lines(errors, warnings):
+    wrapper = TextWrapper(initial_indent='\t\t', subsequent_indent='\t\t', width=80)
+    lines = []
+    if warnings:
+        lines.append('\t%s' % 'Warnings:')
+        first = True
```
vasiliy-bout: We can shrink this method to:
```python
def format_loading_messages_by_lines(errors, warnings):
    wrapper = TextWrapper(initial_indent='\t\t', subsequent_indent='\t\t', width=80)
    lines = []
    if warnings:
        lines.append('\t%s' % 'Warnings:')
        lines.append('\n'.join(['%s\n' % wrapper.fill(warning) for warning in warnings]))
    if errors:
        lines.append('\t%s' % 'Errors:')
        lines.append('\n'.join(['%s\n' % wrapper.fill(error) for error in errors]))
    return lines
```

ChangeLog
```diff
@@ -1,3 +1,7 @@
+* 0.15.1
```
vasiliy-bout: Fix ChangeLog version after merging with `master`

ChangeLog
ekozobrodov: I think it would be better to discuss it with @genestack-solomatin

genestack_client/version.py
```diff
@@ -1,2 +1 @@
 __version__ = '0.16.0a1'
```
klas-genestack: Bump it to **a2**.

genestack_client/scripts/genestack_application_manager.py
ekozobrodov: I tried to implement waiting function with callback and this solution looks to complex and difficult for usage. So, i decided to simply return tuple with success flag and descriptor from `wait_application_loading` function.

PR #213, opened 2017-05-30
#7001 Batch commands should not fail altogether on first error

ekozobrodov:
[#7001](https://trac.genestack.com/ticket/7001)


genestack_client/scripts/genestack_application_manager.py
```diff
@@ -455,15 +455,22 @@ def mark_as_stable(application, version, app_id_list, scope):
...
                 sys.stdout.flush()
-        else:
-            application.invoke('markAsStable', app_id, scope, version)
-            sys.stdout.write('ok\n')
-            sys.stdout.flush()
+        except GenestackServerException as e:
+            if e.debug:
+                raise e
+            else:
+                sys.stdout.write("Cannot mark application '%s' as stable: %s\n" % (app_id, e.message))
```
genestack-solomatin: How do you think, should we need to add something  like 'use --debug  to see traceback'?

ekozobrodov: I slightly changed output to suggest using debug option. Please, take a look.

PR #218, opened 2017-07-12
files_util: Create datasets in 'My datasets' folder by default

alexander-polovtcev:


genestack_client/files_util.py
```diff
@@ -1,27 +1,28 @@
...
     """
     IMPORTED = 'imported'
     CREATED = 'created'
     TEMPORARY = 'temporary'
     UPLOADED = 'uploaded'
+    MY_DATASETS = 'my datasets'
 
 
-class SortOrder:
+class SortOrder(object):
```
genestack-solomatin: How this code related to commit message "files_util: Create datasets in 'My datasets' folder by default"?

PR #223, opened 2017-08-08
Fixed order of checks

ekozobrodov:
This was done to reduce number of calls to server (was found in [#8556](https://trac.genestack.com/ticket/8556))


ekozobrodov:
We also want to not show "was not loaded yet" message if application loading is already failed. I fixed it in a last commit. So, take a look again, please.

PR #225, opened 2017-08-22
Add infinium annotation support, refs #8802

lebedev-genestack:


lebedev-genestack:
Version and ChangeLog was updated in https://github.com/genestack/python-client/pull/224

PR #231, opened 2017-10-27
Sa/cleanup

genestack-solomatin:


genestack_client/chunked_upload.py
```diff
@@ -224,7 +224,8 @@ def __process_chunk(self, chunk):
                 time.sleep(RETRY_INTERVAL)
                 error = str(e)
                 if self.connection.debug:
-                    sys.stderr.write('%s/%s attempt to upload %s failed. Connection error: %s\n' % (attempt + 1, RETRY_ATTEMPTS, chunk, error))
+                    sys.stderr.write('%s/%s attempt to upload %s failed. Connection error: %s\n' % (
+                        attempt + 1, RETRY_ATTEMPTS, chunk, error))
```
klas-genestack: I would format it like this:
```py
                    sys.stderr.write('%s/%s attempt to upload %s failed. Connection error: %s\n'
                                     % (attempt + 1, RETRY_ATTEMPTS, chunk, error))
```
Template on the first line, data to format with on the second.

PR #233, opened 2017-10-31
shell: Adjust help width to terminal width

genestack-solomatin:
I choose 100 as upper bound for help width, because its looks better for me than 120.

### DIFF
--- a/genestack_client/utils.py
+++ b/genestack_client/utils.py
@@ -9,7 +9,7 @@ from genestack_client import GenestackException
 
 def get_terminal_width():
     """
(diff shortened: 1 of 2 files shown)


genestack_client/utils.py
```diff
@@ -1,11 +1,27 @@
...
 import argparse
+import subprocess
 import sys
 
 from genestack_client import GenestackException
 
 
+def get_terminal_width():
+    """
+    Return terminal width.
```
klas-genestack: "Return terminal width in characters (defaults to 80)."

PR #234, opened 2017-10-31
parser: Add check that password is specified only if email is specified

genestack-solomatin:


genestack_client/settings/genestack_user.py
```diff
@@ -97,3 +97,14 @@ def __interactive_login(self, connection):
             except GenestackAuthenticationException:
                 message = 'Your username or password was incorrect for %s. Please try again.' % self.host
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, User) and
+            self.alias == other.alias and
```
klas-genestack: Do we need to check that aliases match?  I mean, it might be the same user with different aliases.

genestack_client/utils.py
```diff
@@ -19,6 +19,14 @@ def isatty():
         return False
 
 
+class GenestackArgumentParser(argparse.ArgumentParser):
+    def parse_known_args(self, args=None, namespace=None):
+        args, argv = super(GenestackArgumentParser, self).parse_known_args(args, namespace)
+        if not args.user and args.pwd:
```
klas-genestack: Let's write `args.pwd and not args.user`, that way it's harder to misunderstand it as `not (args.user and args.pwd)`.

test/test_arguments.py
```diff
@@ -0,0 +1,54 @@
...
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
+from genestack_client import get_user, make_connection_parser
+from genestack_client.settings.genestack_user import User
+
+
+def test_parse_default_user():
+    user = get_user(make_connection_parser().parse_args())
+    expected = User('tester@genestack.com', alias='tester',
+                    host='localhost:8080', password='pwdTester123')
```
klas-genestack: These seem perfect candidates for being defined as constants.

test/test_arguments.py
```diff
@@ -0,0 +1,54 @@
...
+    args = parser.parse_args()
+    expected = Namespace(debug=False, host=None, pwd=None, show_logs=False, user=None)
+    assert args == expected
+
+
+def test_password_without_user(capsys):
+    parser = make_connection_parser()
+
+    with pytest.raises(SystemExit):
+        parser.parse_args(['-p', 'dumy_token'])
```
klas-genestack: "dummy"

klas-genestack: Moreover, let it be something like "useless_password" for clarity

test/test_arguments.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Emulate output of parser.error(message)
+    f = StringIO()
```
klas-genestack: I don't quite follow why you do this voodoo with `StringIO` and `capsys`

genestack_client/settings/genestack_user.py
genestack-solomatin: This class is slightly inner, and not planned to use by API users a lot.  I prefer to leave it as comparing User class instances, to use it in the tests.  If API user create two aliases to same genestack user, it is  not a bug, it it decision of grownup independent person.

test/test_arguments.py
genestack-solomatin: Lets **test** expected values will be as is.

test/test_arguments.py
genestack-solomatin: Because I test `stderr` produced by code using some pytest magic.

PR #236, opened 2017-11-02
Cleanup

genestack-solomatin:
Rename helper functions
Don't pass default user to remove command
Extract common code to separate functions
Remove periods from messages
Make quotes usage more consistent

### DIFF
--- a/genestack_client/scripts/genestack_user_setup.py
+++ b/genestack_client/scripts/genestack_user_setup.py
@@ -25,7 +25,7 @@ def validate_alias(alias):
     return bool(alias and expression.match(alias))
 
 
-def input_alias(existed):
+def input_alias(existing):
     print 'Please input alias. (Alias can contain: letters (a-Z), digit (0-9), at (@), underscore (_), hyphen (-))'
     while True:
         alias = raw_input('alias: ').strip()
@@ -35,7 +35,7 @@ def input_alias(existed):
         if not validate_alias(alias):
             print 'Restricted symbols message'
             continue
-        if alias in existed:
+        if alias in existing:
             print 'Alias must be unique.'
             continue
         return alias
@@ -43,15 +43,14 @@ def input_alias(existed):
 
 def _select(items, message, to_string=None, selected=None):
     """
-    Interactively asks user a message to choose one of the items.
+    Asks user to choose one of the items.
 
-    :param items: list of strings
-    :param message: message
-    :type to_string: function to convert item to string
+    :param items: list of possible choices
+    :param message: clarifying message
+    :type to_string: function to convert choice to string
     :return:
     """
-    if not to_string:
-        to_string = str
+    to_string = to_string or str
 
     if selected and selected not in items:
         raise GenestackException('Selected item "%s" is not in list:%s' % (selected, items))
@@ -76,14 +75,14 @@ def _select(items, message, to_string=None, selected=None):
             print 'Wrong number: "%s".' % raw_index
             continue
 
-        user_index = int(raw_index) - 1
+        item_index = int(raw_index) - 1
 
-        if user_index < 0:
+        if item_index < 0:
             print 'Number is not in list.'
             continue
 
         try:
-            return items[user_index]
+            return items[item_index]
         except IndexError:
             print 'Number is not in list.'
             continue
@@ -145,7 +144,7 @@ class AddUser(Command):
         host = input_host()
         _, user = input_authentication_data(host, alias=alias)
         config.add_user(user)
-        print 'User "%s" created' % user.alias
+        print 'User "%s" has been created' % user.alias
 
 
 def select_user(users, selected=None):
@@ -206,7 +205,7 @@ class SetDefault(Command):
             print 'Set "%s" as default user.' % user.alias
             config.set_default_user(user)
         else:
-            print 'Default user was not changed'
+            print 'Default user has not been changed'
 
 
 class Remove(Command):
@@ -228,7 +227,7 @@ class Remove(Command):
             print 'Cannot delete default user'
             return
         config.remove_user(user)
-        print '"%s" was removed from config' % user.alias
+        print '"%s" has been removed from config' % user.alias
 
 
 class RenameUser(Command):
@@ -303,8 +302,8 @@ class Init(Command):
         group = parser.add_argument_group('command arguments')
         self.update_parser(group)
         group.add_argument('-H', '--host', default=DEFAULT_HOST,
-                           help='server host, use it to make init with different host,'
-                                ' default: %s' % DEFAULT_HOST,
+                           help='Genestack host, '
+                                'change it to connect somewhere else than %s' % DEFAULT_HOST,
                            metavar='<host>')
         return parser
 
@@ -350,7 +349,7 @@ class UserManagement(GenestackShell):
     def set_shell_user(self, args):
         config_path = config.get_settings_file()
         if not os.path.exists(config_path):
-            print 'No config file was found, starting init'
+            print 'No config file was found, creating one interactively'
             self.process_command(Init(), ['--host', args.host or DEFAULT_HOST], False)
             args.host = None  # do not provide host for future use of arguments
 
--- a/genestack_client/scripts/genestack_user_setup.py
+++ b/genestack_client/scripts/genestack_user_setup.py
@@ -6,13 +6,16 @@ import re
 import sys
 from argparse import ArgumentParser
 from getpass import getpass
+from operator import attrgetter
+
+from genestack_client import GenestackException
 
 from genestack_client import GenestackAuthenticationException
 from genestack_client.genestack_shell import Command, GenestackShell
 from genestack_client.settings import DEFAULT_HOST, User, config
 
 
-def ask_host():
+def input_host():
     host = raw_input('host [%s]: ' % DEFAULT_HOST).strip()
     return host or DEFAULT_HOST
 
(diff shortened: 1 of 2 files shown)


genestack_client/scripts/genestack_user_setup.py
```diff
@@ -81,45 +141,25 @@ class AddUser(Command):
...
     def run(self):
-        alias = ask_alias(config.users.keys())
-        host = ask_host()
-        _, user = ask_email_and_password(host, alias=alias)
+        alias = input_alias(config.users.keys())
+        host = input_host()
+        _, user = input_authentication_data(host, alias=alias)
         config.add_user(user)
-        print "User %s created" % user.alias
+        print 'User "%s" created' % user.alias
```
klas-genestack: `'User "%s" has been created'`

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -166,7 +206,7 @@ def run(self):
             print 'Set "%s" as default user.' % user.alias
             config.set_default_user(user)
         else:
-            print "Default user was not changed."
+            print 'Default user was not changed'
```
klas-genestack: "has not been changed"

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -183,12 +223,12 @@ def run(self):
...
         user = users.get(self.args.alias)
         if not user:
-            user = select_user(users, config.default_user)
+            user = select_user(users)
         if user.alias == config.default_user.alias:
             print 'Cannot delete default user'
             return
         config.remove_user(user)
-        print "%s was removed from config" % user.alias
+        print '"%s" was removed from config' % user.alias
```
klas-genestack: "has been removed"

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -260,10 +300,11 @@ class Init(Command):
     def get_command_parser(self, parser=None):
         parser = parser or ArgumentParser(description=self.DESCRIPTION)
         parser.description = self.DESCRIPTION
-        group = parser.add_argument_group("command arguments")
+        group = parser.add_argument_group('command arguments')
         self.update_parser(group)
         group.add_argument('-H', '--host', default=DEFAULT_HOST,
-                           help="server host, use it to make init with different host, default: %s" % DEFAULT_HOST,
+                           help='server host, use it to make init with different host,'
+                                ' default: %s' % DEFAULT_HOST,
```
klas-genestack: That text really baffles me, what does it mean?
"Genestack host, change it to connect somewhere else than $DEFAULT"?

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -278,29 +319,38 @@ def run(self):
...
+        RenameUser
+    ]
     intro = "User setup shell.\nType 'help' for list of available commands.\n\n"
     prompt = 'user_setup> '
 
     def set_shell_user(self, args):
         config_path = config.get_settings_file()
         if not os.path.exists(config_path):
-            print "No config file was found; starting init."
+            print 'No config file was found, starting init'
```
klas-genestack: "starting init" -> "creating one interactively"

genestack_client/settings/genestack_config.py
```diff
@@ -194,5 +168,31 @@ def save(self):
         with open(config_path, 'w') as f:
             document.writexml(f, indent='', addindent='    ', newl='\n')
 
+    def store_secure_value(self, alias, secret_item, key, document, user_element, element_name):
```
klas-genestack: `store_value_securely`

klas-genestack: 1. This need documenting.
2. I would expect arg order and naming like this: `(self, user_alias, key, value, document, user_element, element_name)`.
3. Ideally I would like to have this function accept only `(self, user_alias, key, value)` (only info required for storing in keychain) and returning XML (sub)element (if any) to write into XML config.

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -38,8 +41,64 @@ def ask_alias(existed):
         return alias
 
 
-def ask_email_and_password(host, alias=None):
-    print 'Please input your email and password for %s' % host
+def _select(items, message, to_string=None, selected=None):
+    """
+    Interactively asks user a message to choose one of the items.
+
+    :param items: list of strings
```
klas-genestack: "list of possible choices"

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -38,8 +41,64 @@ def ask_alias(existed):
...
+    :param message: message
+    :type to_string: function to convert item to string
```
klas-genestack: "function to convert choice to string"

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -32,14 +35,69 @@ def ask_alias(existed):
...
+        if not raw_index and selected:
+            return selected
+
+        if not raw_index.isdigit():
+            print 'Wrong number: "%s".' % raw_index
+            continue
+
+        item_index = int(raw_index) - 1
+
+        if item_index < 0:
```
klas-genestack:
```py
if not 0 <= item_index <= (len(items) - 1):
    print('Number is not in list')
```
And drop later `try.. except..`.

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -32,14 +35,69 @@ def ask_alias(existed):
         if not validate_alias(alias):
             print 'Restricted symbols message'
             continue
-        if alias in existed:
+        if alias in existing:
             print 'Alias must be unique.'
```
klas-genestack: Let's consistently *omit* trailing period.

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -32,14 +35,69 @@ def ask_alias(existed):
...
+
+    :param items: list of possible choices
+    :param message: clarifying message
+    :type to_string: function to convert choice to string
+    :return:
+    """
+    to_string = to_string or str
+
+    if selected and selected not in items:
+        raise GenestackException('Selected item "%s" is not in list:%s' % (selected, items))
```
klas-genestack: `in list:_%s`
(insert space)

genestack_client/scripts/genestack_user_setup.py
```diff
@@ -32,14 +35,69 @@ def ask_alias(existed):
...
+
+    default_message = ''
+
+    while True:
+        for i, option in enumerate(items, start=1):
+            if option == selected:
+                print '* ',
+                default_message = ' [%s]' % i
```
klas-genestack: `default_message` -> `about_default`, for example

genestack_client/settings/genestack_config.py
```diff
@@ -194,5 +170,63 @@ def save(self):
...
+        :rtype: bool
+        """
+        try:
+            import keyring
+            keyring.set_password(key, alias, secret_value)
+            return True
+        except Exception as e:
+            if not self.store_raw:
+                sys.stderr.write('Cannot store in secure storage: %s\n' % e)
+            return False
```
klas-genestack: I've just thought: wouldn't raising error from within this function be more logical and Pythonic than returning `bool`?  So instead of
```py
if not self._store_value_securely(user.alias, user.password, GENESTACK_SDK):
    self._store_value_insecurely(user.password, document, user_element, 'password')
```
we could write
```py
try:
    self._store_value_securely(user.alias, user.password, GENESTACK_SDK)
except GenestackException:
    self._store_value_insecurely(user.password, document, user_element, 'password')
```
What do you think?

genestack_client/settings/genestack_config.py
```diff
@@ -194,5 +172,62 @@ def save(self):
...
+            except KeyboardInterrupt:
+                self.store_raw_session = save_to_file
+        if save_to_file:
+            value_element = document.createElement(element_name)
+            value_element.appendChild(document.createTextNode(value))
+            parent.appendChild(value_element)
+        else:
+            sys.stderr.write('"%s" has not been saved to config file\n' % element_name)
+
+    def _store_value_securely(self, alias, secret_value, key):
```
klas-genestack: 1. I think we must use the same argument order as `keyring.set_password()`.
2. These argument names are weird and unhelpful
3. Why do we have `key` here at all?  It's module constant, do we expect to save values into otherwise named *subvaults*, not only "Genestack SDK"?

So I propose:
```py
def _store_value_securely(self, key, value):
    """
    Save value in security vault under ``KEYRING_ENTRY``.
    """
    ...
        keyring.set_password(KEYRING_ENTRY, key, value)
```
Arg description in docstring have to be updated too, obviously.

genestack-solomatin: I want to save tokens and password with different keys.

genestack_client/settings/genestack_config.py
```diff
@@ -49,8 +52,8 @@ def remove_user(self, user):
         self.save()
         try:
             import keyring
-            if keyring.get_password(GENESTACK_SDK, user.alias):
-                keyring.delete_password(GENESTACK_SDK, user.alias)
+            if keyring.get_password(_PASSWORD_KEYRING, user.alias):
+                keyring.delete_password(_PASSWORD_KEYRING, user.alias)
         except ImportError:
```
klas-genestack: Only `import keyring` should be guarded against `ImportError`, and `Exception` could be thrown in get/delete_password only.

PR #237, opened 2017-11-10
Sa/cleanup

genestack-solomatin:


genestack_client/scripts/genestack_user_setup.py
```diff
@@ -114,11 +114,11 @@ def input_email_and_password(host, alias=None):
             continue
         user = User(user_login, host=host, password=user_password, alias=alias)
         try:
-            connection = user.get_connection()
+            user.get_connection()
             break
         except GenestackAuthenticationException:
             print 'Your username or password was incorrect, please try again'
-    return connection, user
+    return user
```
klas-genestack: I dislike how `input_email_and_password` returns `User` object.  How is it different from `login_by_email` conceptually?

genestack-solomatin: It asks user to input email.

klas-genestack: I can see *that*.
I think proper way to do this is instead of calling `input_email_and_password() -> User` to call `input_email_and_password() -> (email, password)`, then call try logging in with `Connection.login(email, password)` or indeed `User()` if we need that object later on.

PR #242, opened 2017-11-21
Add docs entry for DatasetsUtil

genestack-solomatin:


docs/source/api_reference.rst
```diff
@@ -68,6 +68,12 @@ Expression Navigator
         :members:
         :undoc-members:
 
+DatasetsUtil
+------------
+.. automodule:: genestack_client.datasets_util
```
nratskevich: `DatasetsUtil` has only one class, wouldn't it be better to document it as `autoclass`, like `FilesUtil`?

PR #244, opened 2017-12-12
Add token support

genestack-solomatin:


genestack_client/genestack_exceptions.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, current_version, required_version=None):
         :type required_version: distutils.version.StrictVersion
         """
         if required_version:
-            branch = 'master' if required_version.prerelease else 'stable'
+            branch = 'inception' if required_version.prerelease else 'stable'
```
klas-genestack: Huh?  I thought this change should have been rolled back altogether.

PR #245, opened 2017-12-13
Normalization apps was added to client, fix 's' -> 'z' problem (GS-9720) [NOT OT MERGE]

lebedev-genestack:


genestack_client/cla.py
```diff
@@ -191,12 +191,24 @@ def create_merged_reads(self, sources_folder, grouping_key):
 
 # preprocess microarrays
 
-class AffymetrixMicroarraysNormalisationApplication(CLApplication):
```
genestack-solomatin: We cannot remove existing classes. We should deprecate them first. I will try to find example.

genestack_client/expression_navigator.py
```diff
@@ -2,7 +2,7 @@
 
 
 class _BaseExpressionNavigator(Application):
-    def _create_file(self, groups, organism=None, normalised_input=None, options=None):
```
genestack-solomatin: Need to have both keys, for deprecation period.

ChangeLog
```diff
@@ -1,6 +1,7 @@
 * 0.20.0
     * Renamed `password` command  to `change-password` (in `genestack-user-setup`)
     * Added `FileUtils.share_folder` method
+    * Normalization apps was moved into client
```
genestack-solomatin: Added ...
Deprecated  bla-bla,  use bla-bla instead

genestack_client/expression_navigator.py
lebedev-genestack: I've returned this parameter back

genestack_client/cla.py
```diff
@@ -191,10 +191,31 @@ def create_merged_reads(self, sources_folder, grouping_key):
...
+
+class GenePixMicroarraysNormalizationApplication(CLApplication):
+    APPLICATION_ID = 'genestack/genepix-normalization'
+
+
+class InfiniumMicroarraysNormalizationApplication(CLApplication):
+    APPLICATION_ID = 'genestack/infinium-methylation-normalization'
+
+
+# TODO: remove, deprecated. We perform renaming in a scope of global 's' -> 'z' refactoring (american style english).
```
genestack-solomatin: We need to use old class name (to fix usages) but with new file id:

Better way to do it is assign class to variable
```
TODO: Deprecated at {current} will be removed at {current} + 3 ....
AffymetrixMicroarraysNormalisationApplication = AffymetrixMicroarraysNormalizationApplication
....
```

genestack_client/expression_navigator.py
```diff
@@ -15,16 +21,21 @@ def _create_file(self, groups, organism=None, normalised_input=None, options=Non
...
-            'normalisedInputMode': bool(normalised_input),
             'accessionList': [acc for group_id, acc in assignments],
             'groupIdList': map(str, [group_id for group_id, acc in assignments]),
             'organism': organism,
             'groupsNameList': group_names,
             'groupsDescriptionList': group_descriptions,
             'programOptions': options,
         }
+
+        # TODO: reomve this 'if' section after removing the normalised_input argument
```
genestack-solomatin: Don't use s in code, just assign z from it.  Put this check on top of file.
```
if s and z: raise both is not allowed
z = z or s
```

PR #248, opened 2018-01-12
Sa/fix token usage

genestack-solomatin:


genestack_client/settings/genestack_user.py
```diff
@@ -76,7 +76,7 @@ def get_connection(self, interactive=True, debug=False, show_logs=False):
         return connection
 
     def __repr__(self):
-        return "User('%s', alias='%s', host='%s', password='%s')" % (self.email, self.alias, self.host, self.password and '*****')
+        return "User('%s', alias='%s', host='%s', password='%s', token='%s')" % (self.email, self.alias, self.host, self.password and '*****', self.token and '*****')
```
klas-genestack: Way too long.

test/test_arguments.py
```diff
@@ -87,5 +87,14 @@ def test_token_and_user(capsys):
     assert err == expected_output
 
 
+def test_token_with_out_user():
```
klas-genestack: `test_token_without_username`

genestack_client/utils.py
```diff
@@ -89,6 +89,9 @@ def get_user(args=None):
         args = make_connection_parser().parse_args()
 
     alias = args.user
+    if not alias and args.token:
+        return User(email=None, host=args.host, password=None, token=args.token)
```
klas-genestack: Is `alias and args.token` even a valid combination?

genestack-solomatin: No, this is checked in parser.   User is not supposed to be created manually only via this helper method

klas-genestack: But if `alias and args.token` is not valid combination, you should either rely on parser to catch that case, or test for it explicitly.  So the check should look like:
```py
    if args.token:
        return User(email=None, host=args.host, password=None,
                    token=args.token)
```
And if you want to guard against `alias and args.token` here as well as in parser, add an explicit assertion before checking for token.

genestack-solomatin: Lets trust the parser

genestack_client/genestack_connection.py
```diff
@@ -127,7 +128,8 @@ def login(self, email, password):
         self.check_version()
         logged = self.application('genestack/signin').invoke('authenticate', email, password)
         if not logged['authenticated']:
-            raise GenestackAuthenticationException("Fail to login %s" % email)
+            hostname = urlsplit(self.server_url).hostname
+            raise GenestackAuthenticationException("Fail to login with %s to %s" % (email, hostname))
```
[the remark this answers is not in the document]
genestack-solomatin: `http://platform.genestack.org/endpoint/`   does not look good for me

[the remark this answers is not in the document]
klas-genestack: I think we should leave full URL here, so that it would fully match what is saved in config file.  In theory, there could be several frontend instances working on one host even.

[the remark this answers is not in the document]
genestack-solomatin: > I think we should leave full URL here, so that it would fully match what is saved in config file.

We store host in config file.

PR #252, opened 2018-01-16
datasets_util: Add method for retrieving dataset's children (GS-10411)

alexander-polovtcev:
https://genestack.atlassian.net/browse/GS-10411

Do I need to update Python Client version again?


genestack_client/datasets_util.py
```diff
@@ -39,6 +41,24 @@ def get_dataset_size(self, accession):
         """
         return self.invoke('getDatasetSize', accession)
 
+    def get_dataset_children(self, accession):
+        """
+        Return children accessions of the provided dataset.
```
genestack-solomatin: `Return children accessions`  or `generator over dataset's children accessions`?

PR #259, opened 2018-02-28
datasets_util: Add method to create datasets with metainfo

alexander-polovtcev:


alexander-polovtcev:
@genestack-solomatin please check that I haven't messed up with versioning

PR #261, opened 2018-03-01
Allow creating empty datasets

alexander-polovtcev:


alexander-polovtcev:
Because I want to catch potential errors when passing empty lists. Creating empty datasets should be performed intentionally

PR #263, opened 2018-03-13
groups_util: add find_group_by_name method

nratskevich:
GroupsUtil app: https://github.com/genestack/genestack-applications/pull/260


genestack_client/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.22.0a4'
+__version__ = '0.22.0a5'
```
dmitry-at-genestack: This change does not belong to this PR.

genestack_client/groups_util.py
```diff
@@ -0,0 +1,20 @@
...
+from genestack_client import Application
+
+
+class GroupsUtil(Application):
+    APPLICATION_ID = 'genestack/groupsUtil'
+
+    def find_group_by_name(self, name):
+        """
+        Finds group with specified name.
+        If there are no or more than one group with this name, an exception is thrown.
```
dmitry-at-genestack: `If there are no groups or more than one group with this name`

PR #270, opened 2018-04-25
Don't mutate argument

genestack-solomatin:


genestack_client/genestack_metainfo.py
```diff
@@ -317,6 +317,7 @@ def parse_metainfo_from_dict(cls, source_dict):
 
     @staticmethod
     def _parse_scalar_value(source_dict):
+        source_dict = dict(source_dict)
         java_type = source_dict.pop('type')
```
alexander-polovtcev: Why do you actually need to pop the `type`?

ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 0.24.0
+    * Fix `Metainfo.parse_metainfo_from_dict(m)` modifies `m`
```
alexander-polovtcev: Add an issue link instead?

PR #273, opened 2018-05-18
Update setup

genestack-solomatin:


setup.py
```diff
@@ -23,6 +31,16 @@
         ],
     },
     classifiers=[
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Developers',
+        'Natural Language :: English',
```
klas-genestack: It's not applicable here, I would say.

setup.py
```diff
@@ -23,6 +31,16 @@
         ],
     },
     classifiers=[
+        'Development Status :: 5 - Production/Stable',
```
klas-genestack: I don't think 0.x.y version could be classified as "Production/Stable".

PR #275, opened 2018-05-22
setup: refuse to install for Python != 2.7

klas-genestack:
As of now, one could mistakenly install Python Client for Python 3, just to break everything.  It should be forbidden, like this:
```
 % pip install .
Processing /Users/skrattaren/genestack/python-client
genestack-client requires Python '>=2.7,<3' but the running Python is 3.6.5
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ setup(
     long_description=long_descrption,
     long_description_content_type="text/markdown",
     install_requires=['keyring', 'requests', 'pyOpenSSL'],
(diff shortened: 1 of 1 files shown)


alexander-polovtcev:
Can we also remove this check? https://github.com/genestack/python-client/blob/master/genestack_client/__init__.py#L6

klas-genestack:
> Can we also remove this check?

I don't think it hurts.  One can run Python Client directly from repo without installing.

Issue #274, opened 2018-05-21
Metadata retrieved is different from Metadata saved


serverhorror:
Hi,

When saving I'm specifying an integer and what I get back is:

```
        "bi:metadata:label:aNumber": [
            {
                "type": "integer",
                "value": "3"
            }
        ],
```

Note how the actual value is a `str` and not an `int`.

It would be nice to:

* (preferrably) return the actual data type that was used when saving
* provide a method to restore the original data type (otherwise every client has to re-implement restoring the data type)

serverhorror:
Hi,

in short, I'm calling the appropriate `add_....` methods. Here's the method that wraps this:

```
def MakeMetainfo(obj=None, properties=None):
    if obj is None:
        obj = Metainfo()
    TYPEMAP = {
        "int": "add_integer",
        "str": "add_string",
        "unicode": "add_string",
        "float": "add_decimal",
        "bool": "add_boolean",
    }
    # labels: {foo: 1, 2, bar: a, b} -- collections.Mapping
    # tags: [one, two, a] -- collections.MutableSequence
    # provider: val -- basestr
    # site: val -- basestr
    for prop in PROPERTIES:
        values = properties[prop]
        if values is not None:
            # labels: {foo: 1, 2, bar: a, b, zord: True, False}
            # collections.Mapping
            if isinstance(values, collections.Mapping):
                for key in values:  # pylint: disable=E1133
                    for elem in values[key]:  # pylint: disable=E1136
                        if isinstance(elem, (basestring, float, int)):
                            meth = TYPEMAP[type(elem).__name__]
                            meth = getattr(obj, meth)
                            meth("bi:metadata:label:" + key, elem)
                        else:
                            obj.add_string("bi:metadata:" + key, str(elem))
            # tags: [one, two, a] -- collections.MutableSequence
            elif isinstance(values, collections.MutableSequence):
                for elem in values:  # pylint: disable=E1133
                    if isinstance(elem, (basestring, float, int)):  # noqa E509
                        meth = TYPEMAP[type(elem).__name__]
                        meth = getattr(obj, meth)
                        meth("bi:metadata:" + prop, elem)
                    else:
                        obj.add_string("bi:metadata:" + prop, str(elem))
            # provider: val -- basestr
            # site: val -- basestr
            elif isinstance(values, (basestring, float, int)):
                obj.add_string("bi:metadata:" + prop, str(values))
    return obj
```

klas-genestack:
[...]

PR #279, opened 2018-05-25
metainfo: implement `__repr__()` for all `MetainfoScalarValue`s

klas-genestack:
Simple `__repr__` implementations for metainfo values.
```py
mi = genestack_client.Metainfo()
KEY = 'gs:test'
mi.add_integer(KEY, 1)
print(mi.get(KEY))
```
Before:
```
[{'type': 'integer', 'value': '1'}]
```
After:
```
[IntegerValue(1)]
```
References #274.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 * 0.25.0
     * Fixed `Metainfo.parse_metainfo_from_dict(m)` modifies `m` https://github.com/genestack/python-client/issues/269
     * Added 'FilesUtil.get_metainfo_values_as_string_list' method
+    * Implement simple `MetainfoScalarValue.__repr__()` for its subclasses
 
 * 0.24.0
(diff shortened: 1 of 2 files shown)


klas-genestack:
> Should we wrap dicts to `OrderedDict` in `__repr__`?

I don't think it's a good idea.  However, when/if we migrate from subclassing `dict` to wrapping it, we might as well wrap an instance of `OrderedDict` and use its capabilities.  But for now I've opted for using `ppring.pformat()` for dicts.

klas-genestack:
Now:
```
Organization({'city': None,
 'country': 'Russia',
 'department': None,
 'email': None,
 'name': 'Genestack',
 'phone': None,
 'postalCode': None,
 'state': None,
 'street': None,
 'url': None})

IntegerValue(1)

StringValue('ololo')

BooleanValue(True)
```

PR #280, opened 2018-05-28
update docs and readme for publication (GS-11072)

nratskevich:
1. `setup.py` is updated with description, categories and other things
2. Installation instruction is updated
3. Some grammar fixes


developer.md
```diff
@@ -0,0 +1,34 @@
...
+
+   [stable](http://genestack-client.readthedocs.io/en/stable/)
+
+### Installation
+
+   - install library
+     - `pip install genestack-client --upgrade`
+   - setup default user
+     - `genestack-user-setup init`
+
```
klas-genestack: I disagree with moving all these into another file.  **Documentation** and **Installation** sections belong to README.

developer.md
```diff
@@ -0,0 +1,34 @@
...
+  ```
+
+### Generate documentation files
+
+ - install [**sphinx**](http://sphinx-doc.org/) and **sphinx-rtd-theme**
+   - `pip install Sphinx sphinx-rtd-theme`
+ - `cd docs`
+ - `make html`
+
+**Note!** `make` use script `make_shell_docs.py` that generates documentation for shell applications
```
klas-genestack: If you are set on creating separate files for whatever reason (I don't see any need for that whatsoever), call it `README.dev.md`.

nratskevich: Do you think that current readme is ok for external developers? Even `Generate documentation files` part?

nratskevich: Maybe, `readme` should stay the same, but PyPi description should be updated? Because current readme doesn't look suitable for PyPi description, too many internal details.

klas-genestack: Why not?  If we are talking about external _developers_ (not just _users_), then how can they improve docs (do "some gramma fixes", for example ;) without generating documentation locally?  current README is quite short and I see no problems for users/dev in skipping sections they aren't interested in.

klas-genestack: I agree with whole README being unsuitable for PyPI description.  Here's how it's done in _jinja_ project: https://github.com/pallets/jinja/blob/master/setup.py#L41 (and I like this approach).

nratskevich: Good example. Ideally a description should contain a simple usage example, but I can't think of anything like that for this python client.

klas-genestack: I don't think it's suitable, so let's leave it for full documentation.

docs/source/index.rst
```diff
@@ -1,7 +1,7 @@
 Genestack Python Client Library
 ===============================
 
-The Genestack Python Client Library is a Python library that allows you to interact programatically with an instance of the Genestack platform.
+The Genestack Python Client Library is a Python library that allows you to interact programmatically with an instance of the Genestack platform.
 
 **Python 2** supported from version `2.7.5`, **Python 3** is not supported.
```
klas-genestack: While we're _fixing gramma_, make that line
```rst
We support **Python 2** from version *2.7.5* and newer, **Python 3** is not supported.
```
(backticks are n/a here, that's not monospace text in rST)

setup.py
```diff
@@ -2,13 +2,17 @@
 # -*- coding: utf-8 -*-
 
 # noinspection PyUnresolvedReferences
+"""
+Python library that allows you to interact programmatically with an instance of
+the Genestack platform.
+"""
```
klas-genestack: Module docstring should appear right after `coding` comment.

setup.py
```diff
@@ -2,13 +2,17 @@
...
 import setuptools
 
 from distutils.core import setup
 exec(open('genestack_client/version.py').read())
 
 with open('README.md') as f:
-    long_descrption = f.read()
+    long_description = f.read()
```
klas-genestack: This is redundant now.

docs/source/getting_started.rst
```diff
@@ -8,7 +8,7 @@ You can install Genestack Python Client Library using ``pip`` if it is available
 
 .. code-block:: sh
 
-    $ pip install https://github.com/genestack/python-client/archive/stable.zip
+    $ pip install genestack-client --upgrade
```
klas-genestack: See above.

README.md
```diff
@@ -7,7 +7,7 @@
 ### Installation
 
    - install library
-     - `pip install https://github.com/genestack/python-client/archive/stable.zip --upgrade`
+     - `pip install genestack-client --upgrade`
```
klas-genestack: One doesn't need `--upgrade` to install the package, I suggest we remove it for clarity.

PR #281, opened 2018-05-28
Fix date metainfo failure in non-UTC timezones

klas-genestack:
`datetime.fromtimestamp()` returns local datetime for timestamp stored in `DateTimeValue`, while in tests we create and compare it with UTC datetime.

-------
I don't think it's the best solution, but this does fix tests, ad I would like to use it as a starting point for the discussion on how do we use and want to use datetime objects here, because it looks like a mess to me.

### DIFF
--- a/test/test_metainfo.py
+++ b/test/test_metainfo.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python2.7
 # -*- coding: utf-8 -*-
 
+import datetime
 import os
 import sys
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


alexander-polovtcev:
I think that the best solution is to simply return the timestamp as-is (as it is currently done in Java)

klas-genestack:
> I think that the best solution is to simply return the timestamp as-is (as it is currently done in Java)

That's what `get_milliseconds()` is for.

alexander-polovtcev:
Then why do you need this method at all? I think that user can decide whether he needs a UTC date or a local date

klas-genestack:
> Then why do we need this method at all? I think that user can decide whether he needs a UTC date or a local date

1. We should ensure that timestamp is stored reliably.  I have a feeling that nobody really knows what time (local or UTC) is converted to timestamp, and it would be confusing to users.
2. Returning datetime object is nice and Pythonic.  I think that we should always return an _aware_ `datetime` object, by default in UTC timezone, and add an optional argument to `get_datetime(local_time=False)` to return local time.

klas-genestack:
@genestack-solomatin, I've rethought the approach to fixing this test failure.

PR #286, opened 2018-06-04
update incompatible client exception message

nratskevich:
stable branch can be loaded from pypi package

### DIFF
--- a/genestack_client/genestack_exceptions.py
+++ b/genestack_client/genestack_exceptions.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+MASTER_BRANCH = 'https://github.com/genestack/python-client/archive/master.zip'
+PYPI_PACKAGE = 'genestack-client'
 
 
 class GenestackException(Exception):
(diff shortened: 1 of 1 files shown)


genestack_client/genestack_exceptions.py
```diff
@@ -82,13 +82,18 @@ def __init__(self, current_version, required_version=None):
...
 
+        package = 'genestack-client' if branch == 'stable' else (
+            'https://github.com/genestack/python-client/archive/{branch}.zip'
+        ).format(branch=branch)
+
         message += (
             'You can update client with the following command:\n'
-            '    pip install https://github.com/genestack/python-client/archive/{branch}.zip'
-        ).format(branch=branch)
+            '    pip install {package}'
```
klas-genestack: Don't we need `--update` here?

genestack_client/genestack_exceptions.py
```diff
@@ -82,13 +82,18 @@ def __init__(self, current_version, required_version=None):
             message = (
                 'Your Genestack Client version "{current_version}" is too old, '
                 'at least "{required_version}" is required.\n'
-                       ).format(current_version=current_version, required_version=required_version)
+            ).format(current_version=current_version, required_version=required_version)
```
klas-genestack: Please don't move closing brackets to the same level of indentation as the statement.
(Just the same below)

PR #294, opened 2018-08-10
Fix content decoding GSR-935

genestack-solomatin:


genestack_client/genestack_connection.py
```diff
@@ -254,7 +254,7 @@ def __invoke(self, path, post_data, trace=None):
         if trace:
             headers['Genestack-Trace'] = 'true'
         f = self.connection.open(path, post_data, headers=headers)
-        response = Response(json.load(f))
+        response = Response(json.loads(f.read(decode_content=True)))
```
Blizarre: I have found that using `raw` is apparently not really document/recommended: https://github.com/requests/requests/issues/2155#issuecomment-50893660

Is that a problem for us?

genestack-solomatin: I have fixed this.  

`raw` is still used but only in deprecated method.

ekozobrodov:
@klas-genestack this PR is very important for Inception project — we have to use dirty hacks to deploy platform. Please, review it!

genestack_client/genestack_connection.py
```diff
@@ -180,15 +186,56 @@ def open(self, path, data=None, follow=True, headers=None):
...
+        Get response from an URL. The URL is the concatenation of the server URL and ``path``.
+
+        :param path: part of URL that is added to self.server_url
+        :param data: dict of parameters, file-like objects or strings
+        :param follow: should we follow a redirection if any?
+        :param headers: dictionary of additional headers; list of pairs is
+                        supported too until v1.0 (for backward compatibility)
+        :type headers: dict[str, str] | list[tuple[str]]
+        :return: response
+        :rtype: Response
```
klas-genestack:
```python
    def perform_request(self, path, data=None, follow=True, headers=None):
        """
        Perform an HTTP request to Genestack server.

        Connects to remote server and sends ``data`` to an endpoint ``path``
        with additional ``headers``.
  
        :param path: URL path (endpoint) to be used (concatenated with
                     ``self.server_url``).
        :param data: ???
        :param follow: should we follow a redirection (if any)
        :param headers: dictionary of additional headers; list of pairs is
                        supported too until v1.0 (for backward compatibility)
        :type headers: dict[str, str] | list[tuple[str]]
        :return: response from server
        :rtype: Response
```

genestack_client/genestack_connection.py
```diff
@@ -180,15 +186,56 @@ def open(self, path, data=None, follow=True, headers=None):
...
+        """
+        if data is None:
+            data = ''
```
klas-genestack: Why not set `data` to empty string by default instead of `None`?

genestack_client/genestack_connection.py
```diff
@@ -180,15 +186,56 @@ def open(self, path, data=None, follow=True, headers=None):
...
+        elif isinstance(data, dict):
+            data = urllib.urlencode(data)
```
klas-genestack: > :param data: dict of parameters, file-like objects or strings

Doesn't look like anything except strings and dict / list of tuples is supported.

genestack_client/genestack_connection.py
```diff
@@ -253,8 +300,8 @@ def __invoke(self, path, post_data, trace=None):
         headers = {}
         if trace:
             headers['Genestack-Trace'] = 'true'
-        f = self.connection.open(path, post_data, headers=headers)
-        response = Response(json.load(f))
+        # noinspection PyProtectedMember
```
klas-genestack: Why is it needed?

genestack_client/genestack_connection.py
genestack-solomatin: Actually it is:
Dictionary, bytes, or file-like object to send in the body of the :class:`Request`.

Looks like we should remove encoding

PR #297, opened 2018-08-17
python-client: Add constants for sample and study links (ODM-17)

alexander-polovtcev:
TODO: decide whether we need to bump Python Client version


vasiliy-bout:
I think we should bump the version, but not in ClientVersionApplication Java code.

PR #300, opened 2018-09-11
python-client: Add Study Design public API (ODM-79)

alexander-polovtcev:
https://genestack.atlassian.net/browse/ODM-79


genestack_client/study.py
```diff
@@ -0,0 +1,88 @@
...
+        Get information about the given study from the project management point of view.
+
+        :param study_accession: study accession
+        :type study_accession: str
+        :return: dictionary with the following entries:
+
+            * ``studyNumber``: number of study unique for the organization,
+            * ``studyStatus``: current workflow step,
+            * ``approved``: mark of passed curator's review,
+            * ``hasApprovalButton``: mark of *Approve/Disapprove* button visibility in UI,
```
genestack-solomatin: boolean?

genestack_client/study.py
```diff
@@ -0,0 +1,88 @@
...
+            * ``approvalButtonHidden``: *true* if *Approve/Disapprove* button was hidden via
```
genestack-solomatin:
```
hasApprovalButton
approvalButtonHidden
``` 
But we don't have any button in python client.

I suggest to rename this entries. Or maybe replace them with one 3-state (yes, no, hidden)

genestack_client/study.py
```diff
@@ -0,0 +1,88 @@
...
+
+        Status is a name of a project workflow stage as it is going to be displayed in the Study
+        Design UI, e.g. "*Draft*", "*Waiting for review*", etc.
+
+        :param study_accession: study accession
+        :type study_accession: str
+        :param new_status: status name
+        :type new_status: str
+        """
+        self.invoke('setStudyStatus', study_accession, new_status)
```
genestack-solomatin: How about some constants and validation?

genestack_client/study.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def set_study_approval_status(self, study_accession, approved):
```
genestack-solomatin: Study have both status and approval status?

genestack_client/study.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def set_curators_group_name(self, group_name):
+        """
+        Change name of the current organization's curators group.
+
+        Curators group members have the ability to approve/disapprove studies through the UI.
+        However, all users are still able to do that using the :meth:`set_study_approval_status`
+        method.
+
+        :param group_name: name of the curators group. If *None* - curators group will be unset.
```
genestack-solomatin: `If *None* - curators group will be unset.`  it does not looks like responsibility of the method with name`set_curators_group_name`

genestack_client/study.py
alexander-polovtcev: yes

genestack_client/study.py
genestack-solomatin: We can use https://pypi.org/project/enum34/  library  (we already have it on machine, because it is request dependency).   In that case we can expect enum instance and don't have additional validation.

genestack_client/study.py
alexander-polovtcev: > But we don't have any button in python client.

But we do in UI (as the docstring states)

> I suggest to rename this entries

This is not possible, this API is already being used by client scripts through direct invokes

genestack_client/study.py
alexander-polovtcev: All logic is contained within the Java code. Status constants are not applicable here, because it can be any string you want

PR #305, opened 2018-09-28
MetainfoScalarValue: fix UnicodeEncodeError

genestack-solomatin:
https://genestack.atlassian.net/browse/GS-11721


genestack_client/metainfo_scalar_values.py
```diff
@@ -38,7 +38,7 @@ def _xstr(arg):
         :return: string representation of the object
         :rtype: str
         """
-        return str(arg) if arg is not None else None
+        return unicode(arg).encode('utf-8', 'ignore') if arg is not None else None
```
klas-genestack: It's not nearly robust enough:
```python
>>> r'ы'
'\xd1\x8b'
>>> unicode(_)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    unicode(_)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd1 in position 0: ordinal not in range(128)
```

genestack_client/metainfo_scalar_values.py
```diff
@@ -38,7 +38,7 @@ def _xstr(arg):
         :return: string representation of the object
         :rtype: str
         """
-        return str(arg) if arg is not None else None
+        return unicode(arg, 'utf-8').encode('utf-8', 'ignore') if arg is not None else None
```
klas-genestack: Now this method accepts *only* `str` arguments.  If it is intended, docstring should be fixed (probably with exception handling).  But I personally would like to see a more thorough attempt to "cast" `object` to `str`.

ChangeLog
```diff
@@ -4,6 +4,7 @@
     * Fixed error when response log printed without traceback.
     * Added `MetainfoRelatedValueFilter` file filter
     * Addded method `FilesUtil.find_metainfo_related_terms` for loading related terms
+    * Fixed `UnicodeEncodeError` when deserialize `MetainfoScalarValue`
```
klas-genestack:
```
"Fix deserialization of Unicode `MetainfoScalarValue`s"
```

genestack_client/metainfo_scalar_values.py
klas-genestack: I've given it a bit more of thought: your changes are basically a no-op.  If `arg = None`, then you return `None`.  If it's a `str`, you _assume_ it's UTF8-encoded, decode it to `unicode`, then encode back to UTF8.  What's the point?

genestack-solomatin: I see you point. Updated.

klas-genestack:
This change has missed 0.29.0.

PR #306, opened 2018-10-02
add method find_metainfo_related_terms (EDM-306,EDM-204)

dmitry-at-genestack:
Added method `FilesUtil.find_metainfo_related_terms()`. 
See [EDM-306](https://genestack.atlassian.net/browse/EDM-306) and [design document](https://docs.google.com/document/d/1QqyT6w3b0xvCFzXbiRxHEzNkeVI6eKhiZXfSe4gnz_Y/edit?usp=sharing).

Merge after https://github.com/genestack/genestack-applications/pull/448.


genestack_client/files_util.py
```diff
@@ -748,3 +749,72 @@ def create_dataset(self, name, dataset_type, children, parent=None):
...
             return self.invoke('createDataset', parent, name, dataset_type, children)
+
+    def find_metainfo_related_terms(
+            self,
+            file_filter,
+            dictionary_accession,
+            relationship_label,
+            transitive,
+            key,
+            prefix_match=False,
```
alexander-polovtcev: same as in the core: using a enum might be better, because `prefix_match=False` doesn't tell anything by itself and requires consulting the documentation

genestack_client/files_util.py
```diff
@@ -748,3 +749,72 @@ def create_dataset(self, name, dataset_type, children, parent=None):
...
+        :param offset: offset for paging
+        :type offset int
+        :param limit: maximum number of results to return
+        :type limit int
+        :return: list of found dictionary term labels
+        :rtype: list[str]
+        """
+        limit = min(self.MAX_RELATED_TERMS_LIMIT, limit)
+        if offset < 0 or limit < 0:
+            raise GenestackException("Search offset/limit cannot be negative")
```
alexander-polovtcev: Why do you need this check? The platform will check these conditions as well

genestack_client/files_util.py
```diff
@@ -748,3 +749,72 @@ def create_dataset(self, name, dataset_type, children, parent=None):
...
+        "broader term" dictionary relation.
+
+        Search with `transitive==False` considers only directly related terms, with
+        `transitive==true` transitively-related terms are also found.
+
+        See also MetainfoRelatedValueFilter.
+
+        :param file_filter: filter matching files for loading related terms
+        :type file_filter FileFilter
+        :param dictionary_accession: dictionary accession, must reference a valid dictionary
```
alexander-polovtcev: `must reference a valid dictionary` - what does it mean?

genestack_client/files_util.py
```diff
@@ -748,3 +749,72 @@ def create_dataset(self, name, dataset_type, children, parent=None):
...
+
+        Another usage is to find all broader (i.e. more general) dictionary terms for attributes
+        present on available files. If available files contain "Brain" and "Lung" for "Tissue"
+        attribute, "Nervous System" and "Respiratory system" will be found when searching for
+        "broader term" dictionary relation.
+
+        Search with `transitive==False` considers only directly related terms, with
+        `transitive==true` transitively-related terms are also found.
+
+        See also MetainfoRelatedValueFilter.
```
alexander-polovtcev: Use reStructuredText format to reference classes in Python documentation. I would also recommend trying to generate Sphinx documentation to check that docstrings will be formatted correctly in the published documentation (if you aren't doing it already)

PR #308, opened 2018-10-08
add GroupsUtil to API references

nratskevich:


docs/source/api_reference.rst
```diff
@@ -36,6 +36,15 @@ FilesUtil
         :members:
         :show-inheritance:
 
+.. _GroupsUtil:
+
+GroupsUtil
+----------------
```
alexander-polovtcev: If I remember correctly Sphinx doesn't like when the number of underscores does not match header's length. Did you try compiling it?

nratskevich: [Sphinx documentation](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) says that number of underscores can be bigger than the header:
> Section headers (ref) are created by underlining (and optionally overlining) the section title with a punctuation character, at least as long as the text
But it looks bad, so I'll fix it

alexander-polovtcev: maybe I'm wrong and IDEA complained about that, not Sphinx

genestack-solomatin: Sphinx drops warnings to console during `make html` about such cases.

PR #316, opened 2018-11-26
share_util: Remove usages of deprecated API

alexander-polovtcev:


genestack_client/share_util.py
```diff
@@ -76,7 +78,9 @@ def share_files_for_edit(self, file_accessions, group_accession, destination_fol
                will be created if this parameter is equal to `None`.
         :type destination_folder: str
         """
-        self.__share(file_accessions, group_accession, destination_folder, 'shareFilesForEditing')
+        self.safe_share_files(
+            file_accessions, group_accession, [ShareUtil.Permissions.EDIT], destination_folder
```
genestack-solomatin: Do we need to add view permission here?

alexander-polovtcev:
> Do you plan to

Please, continue

genestack_client/share_util.py
alexander-polovtcev: No, these permissions are not the same as in the core, so EDIT includes VIEW as well

genestack-solomatin:
> Please, continue

No.

PR #319, opened 2018-12-03
Update change log

nratskevich:


vasiliy-bout:
@nratskevich I have updated the base branch to next-release, and it looks like we need to rebase this branch.

nratskevich:
@vasiliy-bout I don't get it, why `next-release`? In `next-release` this key is not renamed yet.

vasiliy-bout:
@nratskevich Because I was simply wrong. I somehow thought that FF has already happened :( So everything okay, we do not need to rebase anything. But base branch should be next-release now because of today's FF :)

PR #324, opened 2019-05-29
add timeout for request

efimova-lena0:


genestack_client/genestack_connection.py
```diff
@@ -240,7 +240,7 @@ def perform_request(self, path, data='', follow=True, headers=None):
             _headers.update(headers)
         try:
             response = self.session.post(self.server_url + path, data=data, headers=_headers,
-                                         allow_redirects=follow)
+                                         allow_redirects=follow, timeout=os.environ.get("CLIENT_TIMEOUT", None))
```
genestack-solomatin: This variable should be more Genestack specific: `GENESTACK_CLIENT_TIMEOUT`

Is `os.environ.get("CLIENT_TIMEOUT", None)` an integer?

klas-genestack: In case this question is not sarcastic or something, it's `str` or `None`.

klas-genestack: Once again: too long lines
bild

genestack_client/genestack_connection.py
```diff
@@ -239,8 +239,11 @@ def perform_request(self, path, data='', follow=True, headers=None):
         if headers:
             _headers.update(headers)
         try:
-            response = self.session.post(self.server_url + path, data=data, headers=_headers,
-                                         allow_redirects=follow)
+            response = self.session.post(
+                self.server_url + path, data=data, headers=_headers,
+                allow_redirects=follow,
+                timeout=int(os.environ.get("GENESTACK_CLIENT_TIMEOUT"))
+                if os.environ.get("GENESTACK_CLIENT_TIMEOUT", None) else None)
```
klas-genestack: It's unreadable, badly indented and crashes in case of non-numerical value.
Timeout value should be retrieved from env once, then converted to `int` (or even `float`) with a safeguard against invalid values.

PR #330, opened 2019-10-29
application-manager: return nonzero exit status on error

vasiliy-bout:


genestack_client/scripts/genestack_application_manager.py
```diff
@@ -1,4 +1,4 @@
-#!python
+#!/usr/bin/env python2
```
genestack-solomatin: Please revert it, it is special placeholder that will be replaced with proper shebang after installation with pip.

PR #334, opened 2020-06-11
Support Python 3 (ODM-2544)

klas-genestack:
I should have dropped deprecated methods and classes in a separate PR, but it's a bit hard to disentangle those now =\
So the gist is:
- some long-deprecated stuff has been removed (see Changelog for details);
- lots of changes have been done automatically by _future_ package (e.g., relative module imports, aliases and builtins);
- unicode/bytes, new stdlib modules, `.iter*()` methods and all the usual Python 3 routine;
- apart from changes in actual code, update setup metainfo and documentation.

See ODM-2544 for additional info.


genestack_client/genome_query.py
```diff
@@ -1,3 +1,12 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+from future import standard_library
+standard_library.install_aliases()
+from builtins import str
+from builtins import *
+from builtins import object
```
vasiliy-bout, line 9: Looks like `import *` is enough
```suggestion
from builtins import *
```

There are many other files with the same issue.

genestack_client/metainfo_scalar_values.py
```diff
@@ -1,11 +1,23 @@
...
 import datetime
 import os
 import sys
 from pprint import pformat
-from urlparse import unquote, urlparse
+from urllib.parse import unquote, urlparse
 
 from genestack_client import GenestackException
 
+# TODO: drop this kludge when sopport for Python 2 is over
```
vasiliy-bout, line 18:
```suggestion
# TODO: drop this kludge when support for Python 2 is over
```

setup.py
```diff
@@ -21,8 +20,8 @@
     description='Genestack Python Client Library',
     long_description=__doc__,
     long_description_content_type="text/markdown",
-    install_requires=['keyring', 'requests', 'pyOpenSSL', 'jsonschema', 'pyrsistent==0.16.0'],
-    python_requires='>=2.7,<3',
+    install_requires=['keyring', 'requests', 'pyOpenSSL', 'jsonschema', 'future', 'pyrsistent==0.16.0'],
+    python_requires='>2.7,!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```
vasiliy-bout, line 24:
```suggestion
    python_requires='>2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```

test/test_differential_expression.py
```diff
@@ -95,9 +102,9 @@ def test_get_en_stats(conn):
     query.add_filter(GenomeQuery.Filter.REGULATION, GenomeQuery.Regulation.UP)
 
     result = en.get_differential_expression_stats({EN_TUTORIAL_FILE: query})
-    entries = result.values()[0]
+    entries = list(result.values())[0]
```
vasiliy-bout, line 105: I'm not sure how big this list will be. Maybe we should avoid unnecessary work here?
```suggestion
    entries = next(iter(result.values()))
```

klas-genestack: I guarantee you that the list is exactly 1 item long, but `next(iter(...))` looks better and cooler indeed (`list()` has been added automatically).

vasiliy-bout (approved):

PR #349, opened 2022-11-15
Merge master to the stable for the release

nzorkic-genestack:


genestack_client/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.34.0'
+__version__ = '0.35.0a1'
```
klas-genestack, line 1: Version has to be frozen before merging into "stable" (as per instruction).

klas-genestack (commented):
Update ChangeLog for 0.35.0 as well

klas-genestack (approved):

PR #352, opened 2022-11-30
Merge master to stable for 0.35.1 release

nzorkic-genestack:


genestack_client/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.35.0'
+__version__ = '0.36.0'
```
klas-genestack, line 1: Can we make it "0.35.1"?  Adding newline is very much a minor update.

klas-genestack (changes requested):

PR #366, opened 2023-02-01
[ODM-9057] update genestack user setup doc

dgm90:


ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 0.35.6
+    * Update documentation regarding user setup
+
```
klas-genestack, line 3: I don't think we should release a new version at all.  The changes are 100% in user documentation, so updating package on user's side makes absolutely zero difference.

genestack-okunitsyn:
lgtm!

ChangeLog
klas-genestack: I propose to make doc change with no Python Client release.  We have already released four versions with absolutely zero changes in behaviour.  I am prepared to rebuild the docs manually myself to skip this whole release hassle, so please ditch these version bumps if possible.

genestack-okunitsyn: Ofc you can update documentation manually. As you want. :)

Sorry, but ditch version bumps not possible, it's a preparation for future release process update.

dgm90: Discussed in DM with Oleg: I removed the version bump and Klas can proceed with the manual doc release

klas-genestack (approved):

klas-genestack:
[Done](https://genestack-client.readthedocs.io/en/stable/getting_started.html#configuring-credentials)

PR #370, opened 2023-03-06
ODM-9555 update docs for reference genome upload

olepark-gs:


genestack_client/data_importer.py
```diff
@@ -485,17 +485,17 @@ def create_reference_genome(self,
                                 parent=None,
                                 name=None,
                                 description='',
-                                sequence_urls=None,
+                                sequence_urls=[],
```
klas-genestack, line 488: [Never use a mutable default argiment](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).
Also, why this change at all?

olepark-gs: Fixed. The idea here is that sequence files are no longer used in reference genome related functionality. I updated the docs, removed redundant metadata addition and left the parameter untouched for backwards compatibility. Also added an example script that can be used to add genomes

dgm90 (approved):

genestack_client/data_importer.py
```diff
@@ -508,7 +508,7 @@ def create_reference_genome(self,
         :type name: str
         :param description: experiment description
         :type description: str
-        :param sequence_urls: list urls or local path to sequencing files.
+        :param sequence_urls: list urls or local path to sequencing files. (deprecated, unused)
```
klas-genestack, line 511:
```suggestion
        :param sequence_urls: [OBSOLETE] list urls or local path to sequencing files
```

genestack_client/scripts/upload_reference_genome.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3.9
```
klas-genestack, line 1: Why specific Python version?  Also, Python Client is still supposed to work with Python 2.7, so I don't think it's acceptable to have a Python 3-only script here

klas-genestack (changes requested):
Why `genestack_client/scripts/upload_reference_genome.py` should be in this repo at all?

genestack_client/scripts/upload_reference_genome.py
olepark-gs: I will comment regarding both this and in general why this script is in this repo. We need to simplify the process of reference genome upload to ODM, so I needed to introduce a reusable example of how to upload new genomes. Do you think it doesn't fit this repo? How should it be distributed to clients at all? Maybe they don't need such example?
There is a second reason I put it here - I wanted to reuse it in future to upload genomes at ODM init. Now there's a copy of this script in unified repo, and I want to eventually get rid of it in favour of import from python client

olepark-gs: The version is stated like this, again, because it was just copied as is from the unified repo. I will test it using python 2.7 to make sure it works for that version as well, if it should be compatible with both

genestack-okunitsyn (approved):

klas-genestack (approved):

PR #372, opened 2023-03-08
Support Python 3.11

genestack-okunitsyn:


klas-genestack (changes requested):
Update `classifiers` in `setup.py` as well, then =)

genestack_client/share_util.py
```diff
@@ -13,6 +13,11 @@
 from genestack_client import Application
 
 
+try:
+    collectionsAbc = collections.abc
+except AttributeError:
+    collectionsAbc = collections
```
klas-genestack, line 19: Where is this import used?

genestack-okunitsyn: Autotests and publicDataLoader.

genestack-okunitsyn:
> Update `classifiers` in `setup.py` as well, then =)

done.

tropnikovvl (approved):

klas-genestack (approved):

PR #384, opened 2023-07-12
[ODM-10339] Set bdist_wheel universal

genestack-okunitsyn:


m-smazhevsky:
why do you think it will help?

genestack-okunitsyn:
> why do you think it will help?

I was thinking that issue related to python2 never mind :)

PR #406, opened 2023-12-12
[--ODM-10011--]: share study throw on insufficient permission

viapivov:


natalyatodorova (approved):
I'm glad we decided not introduce a flag)

m-smazhevsky (approved):

PR #407, opened 2023-12-15
ODM-10841: simplify logic to determine odm host

m-smazhevsky:
- localhost check was removed because it's not needed anymore
- use /health endpoint to check server status
- remove cycle to resolve host name, checking just `%s/frontend/health` should be enough

### DIFF
--- a/genestack_client/settings/genestack_user.py
+++ b/genestack_client/settings/genestack_user.py
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
 
+#  Copyright (c) 2011-2023 Genestack Limited
+#  All Rights Reserved
+#  THIS IS UNPUBLISHED PROPRIETARY SOURCE CODE OF GENESTACK LIMITED
+#  The copyright notice above does not evidence any
+#  actual or intended publication of such source code.
+
 from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import division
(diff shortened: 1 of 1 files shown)


ghost (dismissed):

genestack_client/settings/genestack_user.py
```diff
@@ -24,25 +30,16 @@ def _get_server_url(host):
...
-    # compatibility with dev settings
-    # TODO Add code that will do migration in configs and remove this check.
-    if host.startswith('localhost'):
-        return 'http://%s/frontend/endpoint' % host
-
     # default to HTTPS if not explicitly defined
     url_stub = host if has_scheme else 'https://%s' % host
-    # trust provided path if it ends with '/frontend'
     url_stub = url_stub.rstrip('/')
+    # trust provided path if it ends with '/frontend'
```
viapivov, line 30: Why exactly we trust it?

m-smazhevsky: Wasn't invented by me, so my assumption: since host address already ends with /frontend, a client might know the ODM address and some details about ODM urls. It doesn't make any sense to convert the path further, e.g. add '/endpoint' to it; we can just use it as it was provided by client, without additional connection checks.

ghost: Mikhail is correct, the idea was that the path containing `/frontend` is a correct one, or at least we don't have a clear way to guess the correct URL

ghost (approved):

genestack-okunitsyn (approved):

alxndr-bnd:
@m-smazhevsky @genestack-okunitsyn guys, hi! Remove me from the Genestack repos, please. 🙏

tropnikovvl:
> @m-smazhevsky @genestack-okunitsyn guys, hi! Remove me from the Genestack repos, please. 🙏

Hello.
As far as I see you do not have access to the Genestack repositories.
And this repository is public.

alxndr-bnd:
Ok, it is on my side then. Thanks for checking!

PR #408, opened 2023-12-28
ODM-10837: Remove install command from genestack_application_manager

nzorkic-genestack:


viapivov (dismissed):

m-smazhevsky (changes requested):
I think `genestack_application_manager.py` could be completely removed

m-smazhevsky (dismissed):

genestack-okunitsyn:
Shall we remove [genestack-uploader](https://github.com/genestack/python-client/blob/0eff6de493624f357a47c3f3ab0dc29f4ac1c42e/setup.py#L34) entrypoint  too?

As I can see target script doesn't exist.

m-smazhevsky:
> Shall we remove [genestack-uploader](https://github.com/genestack/python-client/blob/0eff6de493624f357a47c3f3ab0dc29f4ac1c42e/setup.py#L34) entrypoint too?
> 
> As I can see target script doesn't exist.

nice catch! Was removed here: https://github.com/genestack/python-client/commit/45b3178af5e249e5524abd63a6cda7877f347f53

m-smazhevsky (changes requested):
why genestack_application_manager.py is empty but not deleted? Also `genestack-uploader` still here

m-smazhevsky (approved):

PR #415, opened 2024-02-01
[ODM-11050] Merge python sdks

genestack-okunitsyn:
In general idea of changes is pretty simple, just transfer code from `auxiliary_scripts` to `genestack_client.scripts`, after that rename `genestack_client` to `odm_sdk`.

Also `tox` was introduced as a test framework that help us to test SDK with different versions of python (for now `3.10`, `3.11`, `3.12`) in parallel.

After merging this PR `auxiliary-scripts` repository can be removed from our CI pipeline and archived.


viapivov (approved):

README.md
```diff
@@ -1,40 +1,35 @@
...
-  ```
-  docs/                             - folder with documentation sources
-  genestack_client/                 - folder with library
-  tests/                            - developer tests, require pytest to run.
-  genestack-user-setup              - script to setup user (installed as script)
-  genestack-shell                   - shell (installed as script)
-  setup.py                          - installation setup for genestack_client library
-  ```
+```shell
+python3 -m pip install odm-sdk
```
m-smazhevsky, line 10: I could not find good examples, so I am not sure: how to name the package? Currently it's odm-sdk for installation and odm_sdk in python code. Shouldn't we keep just one name? 

I found this example: https://github.com/googleapis/google-api-python-client. It's `google-api-python-client` for installation and `googleapiclient` inside python code

setup.py
```diff
@@ -1,49 +1,53 @@
...
+    description='SDK for interacting with the Open Data Manager',
     long_description=__doc__,
     long_description_content_type="text/markdown",
-    python_requires='>=3.7',
+    keywords=['genestack', 'odm', 'import', 'share', 'create',
+              'delete', 'curate', 'genomics', 'api'],
+    include_package_data=True,
+    python_requires='>=3.10',
     entry_points={
         'console_scripts': [
```
m-smazhevsky, line 29: wondering how far we are from creating one entry point with options inside?..

odm_sdk/settings/genestack_config.py
```diff
@@ -5,9 +5,9 @@
 
 import sys
```
m-smazhevsky, line 6: filename odm_sdk_config?

odm_sdk/settings/genestack_config.py
```diff
@@ -5,9 +5,9 @@
...
 
-from genestack_client import GenestackException
-from genestack_client.settings.genestack_user import User
-from genestack_client.utils import ask_confirmation
+from odm_sdk import GenestackException
+from odm_sdk.settings.genestack_user import User
```
m-smazhevsky, line 9: since we are renaming genestack to odm in the library name, possibly we need to do the same in the code? For example genestack_user -> odm_user

m-smazhevsky (commented):
1. Let's think how we can test the changes.
2. Possibly put original genestack_client scripts under `core` folder and aux-scripts under `integration` folder. ATM I don't like name `scripts`
3. We need to update documentation regarding packages to install, console usage and usages in python

README.md
genestack-okunitsyn: > Shouldn't we keep just one name?

It's the same name just in a different case because of python limitations, python module can't be in kebab-case. I've seen this approach in a lot of projects. e.g.
- https://github.com/banglafighter/pweb-cli
- https://github.com/cleanlab/cleanlab-studio
- https://github.com/microsoft/playwright-pytest
- https://github.com/allure-framework/allure-python
- etc.

For me this list could continue infinitely, it's harder to find packages that do not follow this approach rather than vice-versa.

setup.py
genestack-okunitsyn: Honestly, IDK.

For me it's pretty far away taking into account that we should support windows installations for our CLI. At the moment python and pip do it for us.

odm_sdk/settings/genestack_config.py
genestack-okunitsyn: It will lead us to rename [this](https://github.com/genestack/python-client/blob/946e58e8593d3908e1b49cf0c121af72e18299aa/odm_sdk/settings/genestack_config.py#L12-L16) piece of code. And a lot more places. 

I just tried to make the same prefix to have a comfortable work in bash using [Tab] and that's it.

For me, it's better to revert changes in CLI rather than continue with rewriting the whole sdk.

odm_sdk/settings/genestack_config.py
genestack-okunitsyn: Answer in the very next conversation.

genestack-okunitsyn:
> Let's think how we can test the changes.

Complete tests is enough I suppose.

> Possibly put original genestack_client scripts ander core folder and aux-scripts under integration folder. ATM I don't like name scripts

For now, it's just a simple merging(which is blocker for reworking CI and dependencies), I don't think we should enlarge the scope of the task. I think it's better to make a clear structure in the scope of reworking SDK itself but not merging it with other activities.

> We need to update documentation regarding packages to install, console usage and usages in python

For sure, we've created some tasks for documentation in this epic.
- https://genestack.atlassian.net/browse/ODM-11236
- https://genestack.atlassian.net/browse/ODM-11199
- https://genestack.atlassian.net/browse/ODM-11235
- https://genestack.atlassian.net/browse/ODM-11039

README.md
m-smazhevsky: - pweb-cli - a very small project, I don't think we can count it as a valid reference.
- cleanlab-studio - better and indeed, follows the same pattern, but still not really well-known.
- playwright - not valid, it's just playwright in python. 
- allure - not valid, it's just allure in python.

In general I am not against this approach, I just want to collect a little bit more evidence

odm_sdk/settings/genestack_config.py
m-smazhevsky: >revert changes in CLI

do you mean this PR or which changes?

m-smazhevsky:
> Complete tests is enough I suppose.

I think some console commands and scripts are not covered by them, we have to check them manually.

> For now, it's just a simple merging

ok, let's agree on plan and when we can do other activities.

odm_sdk/settings/genestack_config.py
genestack-okunitsyn: I meant [those](https://github.com/genestack/python-client/blob/54d83016dfa645da8c22297a3d72e20cb486a237/setup.py#L30-L31) lines.

README.md
genestack-okunitsyn: Even our generated `odm-api` follows this rule...
A package called `odm-api`, but a python module inside of the package called `odm_api`.

genestack-okunitsyn: For sure, helpful link here:
https://pypi.org/search/?q=cli&page=1

genestack-okunitsyn:
> I think some console commands and scripts are not covered by them, we have to check them manually.

For me it's just wasting time, we move code from one location to another without any changes. If there are problems with moving code, tests for `odm-import-data` (which exists in completes) would be failed. But it's my opinion.

@eeliane @egor-alexandrov WDYT?

egor-alexandrov:
From the looks of it all changes are quite mundane. In test we use several aux scripts not only odm-import. So I would prefer to rely on tests. If you want something to be checked additionally please be more specific. Check everything manually is definitely not the way to go here.

odm_sdk/settings/genestack_config.py
m-smazhevsky: https://genestack.atlassian.net/browse/ODM-11243

odm_sdk/settings/genestack_config.py
m-smazhevsky: https://genestack.atlassian.net/browse/ODM-11243

m-smazhevsky (approved):

PR #431, opened 2024-03-06
ODM-5452 move gene dictionaries to sdk

natalyatodorova:
- The` generate_gene_dictionaries.py` is based on `initialise_raw_gene_dictionary.py`from unified for review it's better to refer to original, because I haven't invented the algorythm. From `gene_dictionary_creation.py`just `get_ensembl_organisms` method has been taken.
- added readme
- upload gene dictionaries in `upload_dictionaries_with_specific_metainfo.py `


genestack-okunitsyn (dismissed):

odm_sdk/scripts/internal/README.md
```diff
@@ -0,0 +1,24 @@
+# README.md
+
+The directory contains scripts that are designed for generating source files for some predefined dictionaries, which are
+uploaded by the initial image. The source files can be found in the `odm-init` S3 bucket. These scripts serve as a
```
m-smazhevsky: initial image -> init image + link to repo

odm_sdk/scripts/internal/README.md
```diff
@@ -0,0 +1,24 @@
...
+backup solution in case the original data is lost.
```
m-smazhevsky:
```suggestion
backup solution in case the original data is lost. They can be also used to get the most recent version of the dictionaries.
```

m-smazhevsky (changes requested):
Nice! Just 2 small comments

odm_sdk/scripts/internal/README.md
natalyatodorova: I decided not apply this suggestion, because we actually are using old ensembl version (published in 2019), which is hardcoded. So `most recent version` might be confusing.

m-smazhevsky (approved):

PR #511, opened 2024-11-14
ODM-11968: Support file upload in import_ODM_data script

nzorkic-genestack:


odm_sdk/scripts/import_ODM_data.py
```diff
@@ -442,12 +445,12 @@ def _async_import(kind, metadata_link, params, data_link=None,
         sys.exit(1)
     url = "{}/{}/job/{}/import/{}/".format(
         params.SERVER, COMMON_URL_PREFIX, params.APP_VERSION, kind.replace('_', '-'))
-    if params.ALLOW_DUPLICATES:
+    if params.ALLOW_DUPLICATES and kind != 'file':
         url += '?allow_dups=true'
-    template_id = params.TEMPLATE_ACCESSION_SUPPLIER()
+    template_id = params.TEMPLATE_ACCESSION_SUPPLIER() if kind != 'file' else None
```
m-smazhevsky, line 450: possibly we should also return an error when user submits file, but with allow_dups/template

nzorkic-genestack: Think it is best to just ignore it this time, because this parameter are also affecting study, so if user provides link to a new study instead of the accession, allow_dups is valid for the study but would throw an error because file is there as well.

m-smazhevsky (approved):

PR #532, opened 2025-03-05
ODM-12424: handling exceptions in delete_study_or_template.py script in accordance with server response

m-smazhevsky:
E.g. showing "Default template could not be deleted from the system. Please, set another template as default and try again." error

### DIFF
--- a/odm_sdk/scripts/delete_study_or_template.py
+++ b/odm_sdk/scripts/delete_study_or_template.py
@@ -6,11 +6,11 @@
 from __future__ import print_function, unicode_literals
 
 import re
+import sys
 
 from odm_sdk import GenestackServerException
-from odm_sdk.utils import make_connection_parser, get_connection
-
 from odm_sdk.scripts.utils import colored, GREEN, RED
+from odm_sdk.utils import make_connection_parser, get_connection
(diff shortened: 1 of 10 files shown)


natalyatodorova (dismissed):

odm_sdk/scripts/delete_study_or_template.py
```diff
@@ -27,13 +26,13 @@ def main():
...
-                          RED))
-        else:
-            print(colored(e, RED))
+        message = e
+        if e.stack_trace is not None:
+            p = re.compile(r"GenestackRestApiException: (.*?)(?=\n)")
+            result = p.search(e.stack_trace)
+            if result:
+                message = result.group(1)
+        print(colored(message, RED))
```
egor-alexandrov: Shouldn't it exit with code 1 in case of an error?

```suggestion
        print(colored(message, RED), file=sys.stderr)
        return sys.exit(1)
```

m-smazhevsky: Seems like we do so in other scripts, so yes, I can add

natalyatodorova (approved):

PR #539, opened 2025-03-27
ODM-12455: when updating a template, always add technical fields

m-smazhevsky:


egor-alexandrov (approved):

natalyatodorova (approved):
Nice tests 👍 
I hope we won't have to repeat the logic in Java later.

PR #580, opened 2025-10-08
ODM-11749 upload cells via import script

marijan-mihaljev:


odm_sdk/scripts/import_ODM_data.py
```diff
@@ -396,6 +399,10 @@ def _prepare_etl_payload(kind, metadata_link=None, template_id=None, data_link=N
                          data_class=None, measurement_separator=None, source=None, study=None):
     ''' Prepare payload to be sent to ETL as parameters '''
     payload = {}
+    # if import cells, job accepts only dataLink parameter, throws an error when's more
```
m-smazhevsky, line 402: what kind of error will be thrown?

odm_sdk/scripts/import_ODM_data.py
```diff
@@ -714,7 +726,13 @@ def link_by_parent(what, accession_to, accession_from, params):
                                         targetId=accession_to),
                              headers=params.headers)
     if response.ok:
-        print("Successfully linked: [{}]".format(what))
+        # TODO remove this check when https://genestack.atlassian.net/browse/ODM-7793 will be fixed
+        # libraries, preparations and cells have
+        # confirmation print in parent function
+        should_print = not (what.startswith('libraries')
```
m-smazhevsky: can we maybe use LIB_PREP_CELL_TAGS here?

odm_sdk/scripts/import_ODM_data.py
```diff
@@ -1313,23 +1404,35 @@ def collect_all_mapping_file_nodes(nodes):
...
     for sample_node in sample_nodes:
-        sample_node['children'] = copy.deepcopy(all_libs_preps)
+        sample_node['children'] = copy.deepcopy(all_libs_preps_cells)
     if mapping_file_nodes:
         sample_node = sample_nodes[0]
-        libs_preps_node = sample_node['children'][0]
-        libs_preps_node['children'] = all_signal_nodes + mapping_file_nodes
+        libs_preps_cell_node = sample_node['children'][0]
+        cell_nodes = []
+        if libs_preps_cell_node['tag'] in {"libraries", "preparations"}:
```
m-smazhevsky: in LIB_PREP_TAGS?

odm_sdk/scripts/import_ODM_data.py
```diff
@@ -950,6 +972,34 @@ def handle_action(self, tag, value, option_string):
     return LibPrepAction
 
 
+def make_cell_action(parser_state):
+    class CellAction(BaseCustomAction):
+        def handle_action(self, tag, value, option_string):
+            sample_node = parser_state.sample_node_list[-1]
+            if sample_node is None:
+                _err("You've provided cell before sample file or accession. Exit!"
```
m-smazhevsky: but it could be also library or preparation, isn't it?

m-smazhevsky (commented):
Not that I understood everything in details, but looks fine in general. Thanks for the work, especially for the test.

odm_sdk/scripts/import_ODM_data.py
marijan-mihaljev: We will get 400 response with unrecognized field message.

odm_sdk/scripts/import_ODM_data.py
marijan-mihaljev: Yes, correct. I was sticking strictly with the check as if we don't find lib or prep, sample is going to be the parent and we're throwing error only if there is no sample. But you're right, it's better to be descriptive than strict :)

m-smazhevsky (approved):

PR #589, opened 2025-11-18
Odm 12867 import cell expressions

marijan-mihaljev:
Implemented import of cell expressions.


marijan-mihaljev:
Closed due to wrong branch name.

